PR #1, opened 2015-09-09
Re-implemented `graphqlerrors` package to be more idiomatic Go code.

sogko:
Hi @chris-ramon 

Really appreciate the work you've put into so far in making a golang graphql library thats closely based on `graphql-js` JS library released recently. (There are other golang-based graphql libraries but most of them were pre-`graphql-js`). Would love to see where this go.

I would like to make a suggestion for this library to be more Go-idiomatic, specifically with error-handling.
It'll make it easier to use it with other Go packages.

Comments appreciated.

### Details

- Previous implementation requires package user to handle errors in non-idiomatic Go, e.g.:
```go
	document, err := Parse(params)
	if err.Error != nil {
		// handle error
	}
```

Generally, most package users would expect a consistent error handling.
```go
	// Parse() returns `error` interface
	document, err := Parse(params)
	if err != nil {
		// handle error
	}
```

This might seem trivial, but has an intangible benefit of not introducing more friction for other devs to adopt the packages.

Reference: https://github.com/golang/go/wiki/Errors

- Implemented other error types based on `graphql-js`
 - Added functions to create different types of GraphQLError (`LocatedError`, `SynxtaxError`)

### Tested

Does not break existing tests.

```bash
$ go build
$ go test github.com/chris-ramon/graphql-go/language/parser github.com/chris-ramon/graphql-go/language/lexer
ok  	github.com/chris-ramon/graphql-go/language/parser   0.005s
ok  	github.com/chris-ramon/graphql-go/language/lexer   	0.008s
```


chris-ramon:
Wow! :star2: thanks for your contribution @sogko !

This is definitely something need to be improved! and I totally agree with what you had done. :+1: 

What do you think about having shorter file naming for files within pkg errors, so instead of having `formatted_error.go` we just have `formatted.go` and so on.

sogko:
Hi @chris-ramon,

Thanks for taking time to look at this PR, appreciate the effort 👍🏻

Yeah I agree with you, it could be less verbose lol. I've updated the PR with those changes.

Btw, I've been working on the `parser` on another branch with the goal of passing `graphql-js` parser tests.
I'll make a separate PR when its ready.

PR #14, opened 2015-09-28
Adds simple extension test to `parser/schema_parser_test`.

chris-ramon:
Adds last missing `simple extension test` to `parser/schema_parser_test` and marks `schema parser` as done on README, we go all test's ported passing from `graphql-js`.


sogko:
Awesome 👍🏻

PR #15, opened 2015-09-29
Removes `github.com/deckarep/golang-set` dependency.

chris-ramon:
Built on top of #13: 

#### Details
- Removes `github.com/deckarep/golang-set` dependency.
- Was being use within `executor/nonnull_test.go` through `testutil.EqualSet`, but we can use instead `reflect.Equal`, sorting slices before comparing equality is a simple solution instead of having this dep.

### DIFF
--- a/errors/sortutil.go
+++ b/errors/sortutil.go
@@ -1,0 +1,30 @@
+package graphqlerrors
+
+import "bytes"
+
+type GQLFormattedErrorSlice []GraphQLFormattedError
+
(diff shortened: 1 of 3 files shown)


sogko:
Alright great! Now theres a way to test for errors 👍🏻

I think its fine to just go ahead and completely remove `testutil.EqualSet()` since nothing else is using it.

chris-ramon:
Great! :+1:  thanks for feedback @sogko, merging this one.

PR #16, opened 2015-09-30
Renames package `types` to `gqltypes`

chris-ramon:
#### Details

Renames `types` pkg name to `gqltypes` so `graphql-go` users can have the `types` pkg name available to use for their programs.

I went ahead a named `gqltypes` instead of `GQLTypes` or `gqlTypes` or something else, because of go's package naming convention, as explained here: [package-names](https://blog.golang.org/package-names).

### DIFF
--- a/executor/abstract_test.go
+++ b/executor/abstract_test.go
@@ -1,13 +1,14 @@
 package executor_test
 
 import (
-	"github.com/chris-ramon/graphql-go"
-	"github.com/chris-ramon/graphql-go/errors"
-	"github.com/chris-ramon/graphql-go/language/location"
-	"github.com/chris-ramon/graphql-go/testutil"
-	"github.com/chris-ramon/graphql-go/types"
 	"reflect"
 	"testing"
+
+	"github.com/chris-ramon/graphql-go"
+	"github.com/chris-ramon/graphql-go/errors"
+	"github.com/chris-ramon/graphql-go/gqltypes"
+	"github.com/chris-ramon/graphql-go/language/location"
+	"github.com/chris-ramon/graphql-go/testutil"
 )
 
 type testDog struct {
@@ -26,38 +27,38 @@ type testHuman struct {
 
 func TestIsTypeOfUsedToResolveRuntimeTypeForInterface(t *testing.T) {
 
-	petType := types.NewGraphQLInterfaceType(types.GraphQLInterfaceTypeConfig{
+	petType := gqltypes.NewGraphQLInterfaceType(gqltypes.GraphQLInterfaceTypeConfig{
 		Name: "Pet",
-		Fields: types.GraphQLFieldConfigMap{
-			"name": &types.GraphQLFieldConfig{
-				Type: types.GraphQLString,
+		Fields: gqltypes.GraphQLFieldConfigMap{
+			"name": &gqltypes.GraphQLFieldConfig{
+				Type: gqltypes.GraphQLString,
 			},
 		},
 	})
 
 	// ie declare that Dog belongs to Pet interface
-	_ = types.NewGraphQLObjectType(types.GraphQLObjectTypeConfig{
+	_ = gqltypes.NewGraphQLObjectType(gqltypes.GraphQLObjectTypeConfig{
 		Name: "Dog",
-		Interfaces: []*types.GraphQLInterfaceType{
+		Interfaces: []*gqltypes.GraphQLInterfaceType{
 			petType,
 		},
-		IsTypeOf: func(value interface{}, info types.GraphQLResolveInfo) bool {
+		IsTypeOf: func(value interface{}, info gqltypes.GraphQLResolveInfo) bool {
 			_, ok := value.(*testDog)
 			return ok
 		},
-		Fields: types.GraphQLFieldConfigMap{
-			"name": &types.GraphQLFieldConfig{
-				Type: types.GraphQLString,
-				Resolve: func(p types.GQLFRParams) interface{} {
+		Fields: gqltypes.GraphQLFieldConfigMap{
+			"name": &gqltypes.GraphQLFieldConfig{
+				Type: gqltypes.GraphQLString,
+				Resolve: func(p gqltypes.GQLFRParams) interface{} {
 					if dog, ok := p.Source.(*testDog); ok {
 						return dog.Name
 					}
 					return nil
 				},
 			},
-			"woofs": &types.GraphQLFieldConfig{
-				Type: types.GraphQLBoolean,
-				Resolve: func(p types.GQLFRParams) interface{} {
+			"woofs": &gqltypes.GraphQLFieldConfig{
+				Type: gqltypes.GraphQLBoolean,
+				Resolve: func(p gqltypes.GQLFRParams) interface{} {
 					if dog, ok := p.Source.(*testDog); ok {
 						return dog.Woofs
 					}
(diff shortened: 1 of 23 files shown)


sogko:
This is great, maybe we should take a closer look into the existing package names exposed in `graphql-go`.
What are your thoughts on the others? Do you have plans for them?

Currently we have:
- `gql`
- `types` => `gqltypes`
- `graphqlerrors`
- `executor`
- `validator`
- Those under `languages` directory
  - `ast`
  - `kind`
  - `lexer`
  - `location`
  - `parser`
  - `printer`
  - `source`
  - `visitor`

----

My own personal thoughts:
- Those packages under `languages` (i.e. `ast`, `parser` etc)  are pretty internal and self-contained, so very unlikely anyone else would want to import them, outside of `graphql-go`
- Same goes with `executor` and `validator`. Very unlikely to be used by other users outside of `graphql-go`
- The most common packages that would be used by other users would be
  - `gql`: Main entry package `gql.Graphql()`
  - `gqltypes`: For schema definition and consuming results from 
     `gql.Graphql(...) *gqltypes.GraphQLResult`
  - `graphqlerrors`: For handling errors from `gql.Graphql()`
     - Possibly rename `graphqlerrors` to `gqlerrors` for consistency with `gqltypes`
- From the perspective of someone who might use `graphl-go` as a library, `gqltypes` is a step up from `types`. It avoids possible package name collision with other libraries.👍🏻
But currently it still suffer from ["stuttering"](https://blog.golang.org/package-names#TOC_3.).
Also, its structs are pretty verbose, which results in a very lengthy schema definition


#### Defining a schema with `graphql-js` (javascript)
```javascript
     var BlogSchema = new GraphQLSchema({
      query: new GraphQLObjectType({
        name: 'Query',
        fields: {
          article: {
            args: { id: { type: GraphQLString } },
            type: BlogArticle
          },
          feed: {
            type: new GraphQLList(BlogArticle)
          }
        }
      })
    });
   graphql(BlogSchema, "query { .... }")
```

#### Defining the same schema with `graphql-go ` (golang)
```go
import (
	"github.com/chris-ramon/graphql-go"
	"github.com/chris-ramon/graphql-go/types"
)
	blogSchema, err := gqltypes.NewGraphQLSchema(gqltypes.GraphQLSchemaConfig{
		Query: gqltypes.NewGraphQLObjectType(gqltypes.GraphQLObjectTypeConfig{
			Name: "Query",
			Fields: gqltypes.GraphQLFieldConfigMap{
				"article": &gqltypes.GraphQLFieldConfig{
					Type: blogArticle,
					Args: gqltypes.GraphQLFieldConfigArgumentMap{
						"id": &gqltypes.GraphQLArgumentConfig{
							Type: gqltypes.GraphQLString,
						},
					},
				},
				"feed": &gqltypes.GraphQLFieldConfig{
					Type: gqltypes.NewGraphQLList(blogArticle),
				},
			},
		}),
	})
       ...
	resultChannel := make(chan *types.GraphQLResult)
	go gql.Graphql(p, resultChannel)
	result := <-resultChannel
	return result
```
I'm proposing that we relook into the existing package names as a whole, so that we can achieve conciseness for consumption by other developers, in particular `gql, gqltypes, graphqlerrors`

### Proposal 1

One possible proposal is to rename GraphQL Types struct names to remove stuttering:
  - e.g `gqltypes.GraphQLSchema` => `gqltypes.Schema`

 ```go
import (
	"github.com/chris-ramon/graphql-go"
	"github.com/chris-ramon/graphql-go/types"
)
	blogSchema, err := gqltypes.NewSchema(gqltypes.SchemaConfig{
		Query: gqltypes.NewObjectType(gqltypes.ObjectTypeConfig{
			Name: "Query",
			Fields: gqltypes.FieldConfigMap{
				"article": &gqltypes.FieldConfig{
					Type: blogArticle,
					Args: gqltypes.FieldConfigArgumentMap{
						"id": &gqltypes.ArgumentConfig{
							Type: gqltypes.String,
						},
					},
				},
				"feed": &gqltypes.FieldConfig{
					Type: gqltypes.NewList(blogArticle),
				},
			},
		}),
	})
	resultChannel := make(chan *gqltypes.Result)
	go gql.Graphql(p, resultChannel)
	result := <-resultChannel
	return result
```

### Proposal 2 
Another possible proposal 
- Bring `gqltypes` into `gql` package
  - e.g `gqltypes.GraphQLSchema` => `gql.GraphQLSchema`
- Rename GraphQL Types struct names to remove stuttering:
  - e.g `gql.GraphQLSchema` => `gql.Schema`
- It is very unlikely that a user would use `gqltypes.*` without `gql.GraphQL()` and vice-versa.
- It would allow for a slightly terse schema definition, for e.g.:

 ```go
import (
	"github.com/chris-ramon/graphql-go"
)
	blogSchema, err := gql.NewSchema(gql.SchemaConfig{
		Query: gql.NewObjectType(gql.ObjectTypeConfig{
			Name: "Query",
			Fields: gql.FieldConfigMap{
				"article": &gql.FieldConfig{
					Type: blogArticle,
					Args: gql.FieldConfigArgumentMap{
						"id": &gql.ArgumentConfig{
							Type: gql.String,
						},
					},
				},
				"feed": &gql.FieldConfig{
					Type: gql.NewList(blogArticle),
				},
			},
		}),
	})
	resultChannel := make(chan *gql.Result)
	go gql.Graphql(p, resultChannel)
	result := <-resultChannel
	return result
```

What are your thoughts?

chris-ramon:
@sogko I def agree on improve the apis that users commonly will use, thanks for proposals and taking the time to look close to this.

Totally agree on moving `types` and `schema` to `graphql-go` pkg like so:
- `graphql-go/types/types.go` -> `graphql-go/types.go`
- `graphql-go/types/schema.go` -> `graphql-go/schema.go`

Also to reduce boilerplate and keep consistency when creating a struct instance, we could update those functions that has `Config` suffix, eg. we could have something like:
```go
gql.NewSchema(&gql.Schema{
})
```
instead of:
```go
gql.NewSchema(gql.SchemaConfig{
})
```

While working on removing `Config` suffix  we will need to update the structs definitions and the way new struct instances are created, meaning we would remove the `YYYConfig` structs, which most of  then share the same properties than `YYY` struct, while doing this we could even have the chance to make those properties that are not to, to be pointers looking forward to memory usage improvements, but I think we could tackle this on new pr's.

So we could end-up with something like this:

```go
import (
"github.com/chris-ramon/graphql-go"
)
blogSchema, err := gql.NewSchema(&gql.Schema{
    Query: gql.NewObjectType(&gql.ObjectType{
        Name: "Query",
        Fields: &gql.FieldMap{
            "article": &gql.Field{
                Type: blogArticle,
                Args: &gql.FieldArgumentMap{
                    "id": &gql.Argument{
                        Type: gql.String,
                    },
                },
            },
            "feed": &gql.Field{
                Type: gql.NewList(blogArticle),
            },
        },
    }),
})
resultChannel := make(chan *gql.Result)
go gql.Graphql(p, resultChannel)
result := <-resultChannel
```

sogko:
Thats awesome! I agree with the direction where this is heading 👍🏻
Those two improvements (replacing `YYYConfig structs` and using pointers for constructor args) would definitely bring benefits in user experience and memory performance.

Certainly looking forward to a cleaner API once we got bulk of the implementation out of the way.
Like you said, I think this will need to be tackled over several commits/PRs.

Cheers!

bbuck:
I really like this idea. Using this library for a very short amount of time it's definitely a bit 'overly' verbose, so to speak. I like merging the common APIs into `gql` as well (not only because it's shorter, but I'd rather have `gql` vs `gqltypes` where `types` doesn't add that much more information and the `gql` easily replaces the `GraphQL` prefix.

Another thought that I had, was very specific to `GQLFRParams` -> `ResolveParams`. It may not be shorter, but it's much more clear what it represents as opposed to `FRParams` where I (honestly) don't know what FR is supposed to be (**f**unction **r**esolve?).

chris-ramon:
Alright! :+1: ... thanks for sharing ur thoughts on this, closing this one and further conversation taking place on this issue: #20 this pr will be cover on new one, details [here](https://github.com/chris-ramon/graphql-go/issues/20#issuecomment-151273936).

PR #21, opened 2015-10-14
Create basic example follow http://graphql.org/docs/getting-started/

thanhliem89dn:
Just port a simple example from JS into GoLang follow http://graphql.org/docs/getting-started/
I hope it will make everyone easy to start to learn about Graphql by using graphql-go and understand how to port js schema sources into Golang


sogko:
Hi @thanhliem89dn 

Thanks for sharing your time to contribute this PR 👍🏻

I think thats a great idea having more examples that others can use to draw parallels with the official docs.

/cc: @chris-ramon: What do you think?

chris-ramon:
Thanks a lot @thanhliem89dn! :star2: 

I've tried locally and `get` seems working as expected but not `post` test.

```bash
Test with Get: http://localhost:8080/graphql?query={user(id:%221%22){name}}
```

screen shot 2015-10-18 at 18 39 07

***

```bash
Test with Post: curl -XPOST http://localhost:8080/graphql -H 'Content-Type: application/graphql'  -d 'query Root{ user(id:'1'){name}  }'
```

```bash
{
  "data": {
    "user": null
  }
}
```

Also it think we might want to pluralize `example` -> `examples`, so we can include other examples in this directory as well.

thanhliem89dn:
Thanks for your time to take a look on this pull request. 

Replace '1' in [curl -XPOST http://localhost:8080/graphql -H 'Content-Type: application/graphql'  -d 'query Root{ user(id:'1'){name}  }'] with "1" 
===> curl -XPOST http://localhost:8080/graphql -H 'Content-Type: application/graphql' -d 'query Root{ user(id:"1"){name}  }'

It'll be ok :). And I also fixed curl command in example in new commit beside changed example to examples.

chris-ramon:
Great! :star2:  - thanks for updates @thanhliem89dn 

Just tried and both test are working as expected:

```bash
Test with Get: http://localhost:8080/graphql?query={user(id:%221%22){name}}
```
screen shot 2015-10-19 at 01 52 16
***
```bash
curl -XPOST http://localhost:8080/graphql -H 'Content-Type: application/graphql'  -d 'query Root{ user(id:"1"){name}  }'
```
```bash
{
  "data": {
    "user": {
      "name": "Dan"
    }
  }
}
```
***

Also just for consistency could we use `curl` for `get` as well, like:

```bash
curl -g "http://localhost:8080/graphql?query={user(id:%221%22){name}}"
```

```
{
  "data": {
    "user": {
      "name": "Dan"
    }
  }
}
```

Finally what do you think if we go for snake case naming so `SimpleUserGraph` -> `simple_user_graph` and rename `simpleUserGraph.go` -> `main.go`

cc/ @sogko

thanhliem89dn:
In my opinion, I think camelCase is better snake_case. About rename simpleUserGraph.go -> main.go, It's a good idea :)

chris-ramon:
Alright, so I think we could do the following:

- [ ] Rename `example` -> `examples` - to place other examples within this directory.
- [ ] Rename `simpleUserGraph.go` -> `main.go` - to make it clear where program begins.

Finally I think we could rename `SimpleUserGraph` to `http`, so this one becomes the http example.

cc/ @sogko

thanhliem89dn:
Can't agree more with you. I have changed all of them. Please take a look @chris-ramon @sogko

chris-ramon:
Great! thanks @thanhliem89dn :star2: - to keep consistency could we use `curl` for `get` test as well?.
So we can have something like:

```bash
go run main.go 
Now server is running on port 8080
Test with Post: curl -XPOST http://localhost:8080/graphql -H 'Content-Type: application/graphql'  -d 'query Root{ user(id:"1"){name}  }'
Test with Get: curl -g "http://localhost:8080/graphql?query={user(id:%221%22){name}}"
```

sogko:
Thanks for spending time to contribute this @thanhliem89dn, this looks great!

thanhliem89dn:
thank for your time @chris-ramon, I have just edited curl command for Get example. I think it's nice. One more thing, Why don't We open a git room for communicate :)

chris-ramon:
Alright, thanks again @thanhliem89dn! - yes! good idea to have a chat room, I think we could start by using gitter, [here](https://github.com/chris-ramon/graphql-go/pull/26) the PR.

I would like to share my final thoughts about this, merging this PR will make `graphql-go` depend on `sogko/graphql-go-handler`, when should be the opposite?.

Perhaps this example should be on `sogko/graphql-go-handler` and we can create a link reference within `graphql-go's` README?.

Other solution could be merge both projects `graphql-go` and `graphql-go-handler`, which for me seems ideal since we will always want use a http handler for our graphql schemas, and we could expose a constructor like:
```go
handler := gql.NewHandler()
```
As @sogko pointed out [here](https://github.com/chris-ramon/graphql-go/issues/20#issuecomment-148574465).

Please do share your thoughts about this one, thanks! :+1:

thanhliem89dn:
Thanks for your time to reply,

But I think We shouldn't merge both projects [graphql-go] and [graphql-go-handler]. In my point of view, We almost use http handler but use my own, so keep [graphql-go] separate to [graphql-go-handler] or [relay] is better. So We can use it in any purpose also implement in any old project.

About my example I think It's a very first look for everyone to interact with [graphql-go] as GraphQL official document. But As you told It used [graphql-go-handler] to handle request so I think it's better if I take [graphql-go-handler] away.

chris-ramon:
Alright, yeah I agree @thanhliem89dn, perhaps we should preserve `graphql-go` and `graphql-go-handler` as separated libs by now, in that case we could to the following:

1. Have a reference on `graphql-go` README to a new repository that you would need to create, something like: `thanhliem89dn/graphql-go-examples/http`.

2. Remove `graphql-go-handler` dependency and use plain `net/http`.

thanhliem89dn:
Hi @chris-ramon, I have just taken [graphql-go-handler] out of example and just use net/http. I think it's clear for anyone to start with this RP.

chris-ramon:
Thanks for the updates @thanhliem89dn ! - I've tested locally and is working good, I think is ready to go.

/cc: @sogko

sogko:
I've just tested it on my local machine as well, looks great! 

Thanks @thanhliem89dn! Looking forward to more contributions from you 👍🏻

base698:
How do you parse a schema?  In the node version you pass a string and get a schema object that you can then initialize as a middleware?

PR #23, opened 2015-10-15
Adjust scalar value coersion methods to support the various int types.

### LINKED ISSUE #22: Returning a `uint` in a Resolve function for a `GraphQLNonNull(GraphQLInt)` field responds with `0`
I have a simple test server set up to run off todos in a database. I'm using the gorm ORM and it's default `gorm.Model` that defines common fields for models. One of which happens to be the `ID` field which is of type `uint`. 

```go
todoType := graphql.NewObject(graphql.ObjectConfig{
        Name: "Todo",
        Fields: types.Fields{
                "id": &types.Field{
                        Type: graphql.NewNonNull(graphql.Int),
                                todo := params.Source.(*data.Todo)
                                        
                                return todo.ID
                        },
                },
                // Shortened for brevity
        },
})
```

When I make a query to this, I get back `"id": 0` for all records. If I change the return to `int(todo.ID)` then I get back the expected ID value.

It appears the reason is that `coerceInt` does not support the `uint` type. Also missing are `int8`, `int16`, `int32`, `int64`, and the uint series as well. When I get the opportunity I'll see about making a pull request to address this.

bbuck:
This should resolve #22


sogko:
Hi @bbuck 

Sorry for not been giving attention to this earlier. I think issue #22 and #23 is still valid. 
Would re-look into this now that the codebase is relatively more stable.

Cheers!

bbuck:
I'm currently reimplementing my fix for #22 in the new changes.

sogko:
Awesome, looking forward to that 👍🏻

Thanks again!

PR #36, opened 2015-10-30
golint updates for examples/http

chris-ramon:
#### Details

- Adds updates for the following `golint` advices:

`cd examples/http/`
`$GOPATH/bin/golint main.go `

```bash
main.go:13:6: exported type User should have comment or be unexported
main.go:14:2: struct field Id should be ID
main.go:64:13: if block ends with a return statement, so drop this else and outdent its block
main.go:106:6: func importJsonDataFromFile should be importJSONDataFromFile
```

### DIFF
--- a/examples/http/main.go
+++ b/examples/http/main.go
@@ -10,12 +10,12 @@ import (
 	"github.com/chris-ramon/graphql-go/types"
 )
 
-type User struct {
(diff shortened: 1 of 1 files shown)


sogko:
This is a great way to start using `golint` for the project 👍🏻

PR #39, opened 2015-10-30
Port AllowsEarlyExitWhileVisiting

EmergentBehavior:
Port from graphql-js.

Split from https://github.com/chris-ramon/graphql-go/pull/32.

cc @sogko @chris-ramon


chris-ramon:
Thanks! - looks good to me :+1:

PR #48, opened 2015-11-04
Make executor synchronous

Fugiman:
This changes the API for `graphql.Graphql ` to no longer require a channel as a parameter, and instead just returns the `*Result`. My hope is that this makes the API simpler and improves the readability of the code by reducing the duplication of error handling.

As far as I can tell, the channel was only ever used once, and wasn't required for any sort of concurrency. All the tests still pass. However, it may still be required for future concurrency work. My hope is that if that's the case, the internal function can handle creating the channel itself.

### DIFF
--- a/abstract_test.go
+++ b/abstract_test.go
@@ -145,13 +145,10 @@ func TestIsTypeOfUsedToResolveRuntimeTypeForInterface(t *testing.T) {
 		Errors: nil,
 	}
 
-	resultChannel := make(chan *graphql.Result)
-
-	go graphql.Graphql(graphql.Params{
+	result := graphql.Graphql(graphql.Params{
 		Schema:        schema,
 		RequestString: query,
-	}, resultChannel)
-	result := <-resultChannel
+	})
 	if len(result.Errors) != 0 {
 		t.Fatalf("wrong result, unexpected errors: %v", result.Errors)
 	}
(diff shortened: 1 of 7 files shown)


sogko:
Hi @Fugiman 

Thanks for looking into this, appreciate it very much 👍🏻

Yeah I agree with your observation regarding using channel for results for `graphql.Graphql()`, as far as public API goes.

Most of the time, I end up wrapping `graphql.Graphql()` within a function that deals with the results channel.

```go
func gql(p graphql.Params, ...) {
	resultChannel := make(chan *graphql.Result)
	go graphql.Graphql(p, resultChannel)
	result := <-resultChannel
	...
}
func main() {
	...
	gql(p)
}

```

Internally, one place I remember where concurrency would be useful is the `func executeFields(p ExecuteFieldsParams)` inside `executor` (which is used for executing fields for `read` mode or `query`, [which the specs allow out-of-order execution](http://facebook.github.io/graphql/#sec-Normal-evaluation))

For future work, probably the executor could concurrently resolve fields for queries (i.e. not mutation requests).

/cc @chris-ramon

Fugiman:
Thanks for the feedback. You're right that `executeFields` could use concurrency, but I think the channel could be created in that function rather than by the end-user.

chris-ramon:
Thanks for going ahead and fix this @Fugiman! :star2: 

Initially, I went ahead and use channels because in `graphql-js`'s implementation the `graphql()` function returns a `Promise`, so I just tried to mimic the implementation and looking forward to improve it later.

So I totally agree on making `graphql.Graphl` sync and looking forward to explore optimizations internally with the usage of coroutines.

PR #54, opened 2015-11-05
Add utils package for printing schemas

pyros2097:
The schema printer is in the works

### DIFF
--- a/utils/introspection_query.go
+++ b/utils/introspection_query.go
@@ -1,0 +1,78 @@
+package utils
+
+var IntrospectionQuery = `
+  query IntrospectionQuery {
+    __schema {
+      queryType { name }
+      mutationType { name }
+      subscriptionType { name }
+      types {
+        ...FullType
+      }
+      directives {
+        name
+        description
+        args {
+          ...InputValue
+        }
+        onOperation
+        onFragment
+        onField
+      }
+    }
+  }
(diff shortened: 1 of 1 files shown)


pyros2097:
Oops .. Forgot to ask .. Is the utilities package required? I just want it to be sync with graphql-js. And the package name utils seems better.

sogko:
Hi @pyros2097 

Thanks for volunteering your time and taking this up 👍🏻

Regarding the utilities module from [`graphql-js` perspective,](https://github.com/graphql/graphql-js/blob/master/src/utilities/README.md) it's meant to be a mixed bag of helpers that can be used to other stuff around graphql (for e.g for tooling etc) 

Currently in `graphql-go`, while there isn't a 'util' package per se, some of the utility functions from `graphql-js` has already been ported but it exists in all over. 

(I wasn't sure about having a `util` package at the time of writing, which kind goes against golang distaste for utilities package. Again, I decided to defer the decision to the community and consolidate the util funcs into a better designed package)

Just a quick head up before you go deeper: 
- the `IntrospectionQuery` already exists in the current `testutil` package in `testutil/introspection_query.go`
- `astFromValue()` method that you might need for porting `schemaPrinter` already exists in `introspection.go`.

Cheers!

pyros2097:
This was not needed. Closing it for now.

PR #55, opened 2015-11-05
Update grammar in README.md.

EmergentBehavior:
Some miscellaneous fixes.

I'm also wondering if we should instruct folks to install the package via the corresponding `gopkg.in` URL (for pre-release software, would be using v0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,13 @@
 A *work-in-progress* implementation of GraphQL for Go.
 
 ### Getting Started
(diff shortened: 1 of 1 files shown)


chris-ramon:
LGTM :+1: 

Thanks! for mentioning about pkg versioning, I think we could have the conversation on an issue, I went ahead and created it -> https://github.com/graphql-go/graphql/issues/56

Issue #44, opened 2015-11-03
API Improvement.


chris-ramon:
Before releasing the first version of the lib, I think we could improve the API, meaning remove verbosity and clean-it-up to be go-idiomatic, so please do share ur thoughts on this! :+1: 

This is the basic example from the `Getting Started` section on `README` (still to merge #43):

``` go
package main

import (
    "encoding/json"
    "fmt"
    "log"

    "github.com/chris-ramon/graphql"
)

func main() {
    // Schema
    fields := graphql.FieldConfigMap{
        "hello": &graphql.FieldConfig{
            Type: graphql.String,
            Resolve: func(p graphql.GQLFRParams) interface{} {
                return "world"
            },
        },
    }
    rootQuery := graphql.ObjectConfig{Name: "RootQuery", Fields: fields}
    schemaConfig := graphql.SchemaConfig{Query: graphql.NewObject(rootQuery)}
    schema, err := graphql.NewSchema(schemaConfig)
    if err != nil {
        log.Fatalf("failed to create new schema, error: %v", err)
    }

    // Query
    query := `{hello}`
    params := graphql.Params{Schema: schema, RequestString: query}
    result := make(chan *graphql.Result)
    go graphql.Graphql(params, result)
    r := <-result
    if len(r.Errors) > 0 {
        log.Fatalf("failed to execute graphql operation, errors: %+v", r.Errors)
    }
}
```

My initial proposal is:

``` go
package main

import (
    "encoding/json"
    "fmt"
    "log"

    "github.com/chris-ramon/graphql"
)

func main() {
    // Schema
    helloResolve := func(p graphql.ResolveParams) interface{} {
        return "world"
    }
    fields := []graphql.Fields{
        graphql.Field{Name: "hello", Type: graphql.String, Resolve: helloResolve},
    }
    schema, err := graphql.NewSchema(Name: "RootQuery", Fields: fields)
    if err != nil {
        log.Fatalf("failed to create new schema, error: %v", err)
    }

    // Query
    query := `{hello}`
    result := make(chan *graphql.Result)
    go graphql.Do(schema, query, result) // Like: http.Client.Do
    r := <-result
    if len(r.Errors) > 0 {
        log.Fatalf("failed to execute graphql operation, errors: %+v", r.Errors)
    }
}
```

sogko:
Woah the improvements starting to look good 👍🏻

My humble notes:

##### 1. Resolve function
[...]

PR #62, opened 2015-11-06
Implement graphql.Do as described in #44

Fugiman:
This is an attempt to simplify `graphql.Do` as described in #44. While I greatly enjoy not having to pass in a struct, this appears to remove ability for the user to define `RootValue`, `OperationName` and `VariableValues`. I'm not sure if the struct or just having five arguments is better. Maybe compromise with having a wrapper around `Do` that only takes two arguments for the simple case? I don't know how common it'll be in practice.

This is expected to not pass tests, as I haven't yet updated all the tests to the new API. I'm holding off on that until it's determined what API is ideal.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -505,7 +505,7 @@ type ResolveInfo struct {
 	FieldASTs      []*ast.Field
 	ReturnType     Output
 	ParentType     Composite
-	Schema         Schema
+	Schema         *Schema
 	Fragments      map[string]ast.Definition
 	RootValue      interface{}
 	Operation      ast.Definition
--- a/examples/hello-world/main.go
+++ b/examples/hello-world/main.go
@@ -31,8 +31,7 @@ func main() {
 			hello
 		}
 	`
-	params := graphql.Params{Schema: schema, RequestString: query}
-	r := graphql.Graphql(params)
+	r := graphql.Do(schema, query)
 	if len(r.Errors) > 0 {
 		log.Fatalf("failed to execute graphql operation, errors: %+v", r.Errors)
 	}
--- a/examples/http/main.go
+++ b/examples/http/main.go
@@ -73,11 +73,8 @@ var schema, _ = graphql.NewSchema(
 	},
 )
 
-func executeQuery(query string, schema graphql.Schema) *graphql.Result {
-	result := graphql.Graphql(graphql.Params{
-		Schema:        schema,
-		RequestString: query,
-	})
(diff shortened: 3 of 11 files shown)


sogko:
Hi @Fugiman 

Thanks for working on this PR 👍🏻

Having `graphql.Do(schema, query) does make it look very attractive to use.

I think we still need to pass in `graphql.Params`, because it might not be enough to just have `schema` and `query`. 
There are still valid cases for passing in `root value`, `variable values` and `operation name`.

I wish Go has named arguments right now, or at least remove the need to explicitly write the struct type for functions, similar to map literals. (There are a [couple of proposals](https://github.com/golang/go/issues/12296) [floating around](https://github.com/golang/go/issues/12854), but it won't be here anytime soon)

Writing all five arguments may not be the best for a couple of reasons:
- Only `schema` and `request` are required, the others are optional. This may lead to `graphql.Do(schema, query, nil, nil, nil)` I don't know if thats any better
- Adding a new argument would essentially be very disruptive and break compatibility very easily, as opposed to just added a field to the Params struct. 

But your suggestion of creating a wrapper does sound reasonable. Though I can't think of a good name for it right now lol.

chris-ramon:
I would like to mention that we might consider other possible names for `graphql.Graphl` function, I've consider the following:

1. `graphql.Do`
2. `graphql.Run`
3. `graphql.Exec`
4. `graphql.Execute`

`graphql.Graphql` in summary does:  Source -> Parse -> AST -> Validation -> Execute -> Result

Said that, multiple steps and components are involved, so perhaps `graphql.Execute` and `graphql.Exec` might not be appropriate.

`graphql.Run` - gives the notion of continuos execution.
`graphql.Do` - gives the notion of sync execution.

`graphql.Graphql` was recently [updated](https://github.com/graphql-go/graphql/pull/48) to be sync, so renaming to `graphql.Do` seems to be convenient plus is go-idiomatic.

So please do share ur thoughts on this.

***

How other GraphQL implementations are naming `graphql.Graphql`:
- `github.com/rmosolgo/graphql-ruby` -> `Schema.execute`
- `github.com/sangria-graphql/sangria` -> `Executor.execute`
- `github.com/graphql/graphql-js` -> `graphql`
- `github.com/graphql-python/graphene` -> `schema.execute`
- `github.com/andimarek/graphql-java` -> `GraphQL.execute`

Fugiman:
Closing in favor of #68

PR #69, opened 2015-11-08
Go vet

Fugiman:
`go vet` is a pretty cool tool. I'd highly recommend using it, and requiring code to pass it.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -594,10 +594,6 @@ func completeValue(eCtx *ExecutionContext, returnType Type, fieldASTs []*ast.Fie
 	// If field type is Scalar or Enum, serialize to a valid value, returning
 	// null if serialization is not possible.
 	if returnType, ok := returnType.(*Scalar); ok {
-		err := invariant(returnType.Serialize != nil, "Missing serialize method on type")
-		if err != nil {
-			panic(gqlerrors.FormatError(err))
-		}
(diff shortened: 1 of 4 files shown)


sogko:
Nice! This is great 👍🏻

I add this to the draft contribution guideline.

/cc @chris-ramon

EmergentBehavior:
Worth running `golint` too while you're at it.

Fugiman:
I'm filing separate branches for each tool. Partially to allow for discussion on each one in particular, and partially because `golint` requires a lot of work to have it start passing :)

executor.go
```diff
@@ -594,10 +594,6 @@ func completeValue(eCtx *ExecutionContext, returnType Type, fieldASTs []*ast.Fie
 	// If field type is Scalar or Enum, serialize to a valid value, returning
 	// null if serialization is not possible.
 	if returnType, ok := returnType.(*Scalar); ok {
-		err := invariant(returnType.Serialize != nil, "Missing serialize method on type")
```
chris-ramon: I was wondering about this code elimination, but it does totally make sense :+1:  ... we would get compile error if we would make`Scalar.Serialize ` or `Enum.Serialize` actually `nil`.

chris-ramon:
Thanks! @Fugiman :star2: 

Looks good to me.

chris-ramon:
:+

PR #68, opened 2015-11-08
Renames

Fugiman:
As discussed in #20 and #44, this renames a bunch of structs and methods. I simply used [`gorename`](https://godoc.org/golang.org/x/tools/cmd/gorename) to do this, there are no actual code changes.

`Graphql` -> `Do`
`FieldConfigMap` -> `Fields`
`FieldConfig` -> `Field`
`GQLFRParams` -> `ResolveParams`
`GQLFormattedErrorSlice` -> `FormattedErrors`
`Name` -> `PrivateName`
`Description` -> `PrivateDescription`
`GetName` -> `Name`
`GetDescription` -> `Description`
`GetError` -> `Error`
`GetValues` -> `Values`
`GetFields` -> `Fields`
`GetObjectType` -> `ObjectType`
`GetPossibleTypes` -> `PossibleTypes`
`GetInterfaces` -> `Interfaces`
`GetDirectives` -> `Directives`
`GetMutationType` -> `MutationType`
`GetQueryType` -> `QueryType`
`GetType` -> `Type`
`GetTypeMap` -> `TypeMap`

I couldn't use `name` and `description` due to `encoding/json` only working on exported fields.

PR #73, opened 2015-11-13
Enable cyclic types in ObjectType and InputObjectType

atrniv:
I've implemented the ability to allow the definition of cyclic field configuration. However I'm not too sure about the implementation style since the code seems to implement it in different ways for object, interface and input object. I've also added a simple test cases for the cyclic type in object field configurations. Let me know if there is something else I need to do.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -297,14 +297,15 @@ type Object struct {
 
 type IsTypeOfFn func(value interface{}, info ResolveInfo) bool
 
+type ObjectFieldMapThunk func() FieldConfigMap
 type InterfacesThunk func() []*Interface
 
 type ObjectConfig struct {
-	Name        string         `json:"description"`
-	Interfaces  interface{}    `json:"interfaces"`
-	Fields      FieldConfigMap `json:"fields"`
-	IsTypeOf    IsTypeOfFn     `json:"isTypeOf"`
-	Description string         `json:"description"`
+	Name        string      `json:"description"`
+	Interfaces  interface{} `json:"interfaces"`
+	Fields      interface{} `json:"fields"`
+	IsTypeOf    IsTypeOfFn  `json:"isTypeOf"`
+	Description string      `json:"description"`
 }
 
 func NewObject(config ObjectConfig) *Object {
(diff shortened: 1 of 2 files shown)


sogko:
Hi @atrniv 

Thanks for taking time to contribute to the project, appreciate it very much 👍🏻

You are right, currently the way those three GraphQL Types handle cyclic field definition are different:
- `graphql.Object` and  `graphql.Interface` has a `AddFieldConfig()`
- `graphql.InputObject` uses a `Thunk` (ported from `graphql-js`)

To give a little background, `go` does not like cyclic declaration. So, the following object definition does not work:

```go
func main() {
	// This won't work
	personType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Person",
		Fields: graphql.FieldConfigMap{
			"name": &graphql.FieldConfig{
				Type: graphql.String,
			},
			"bestFriend": &graphql.FieldConfig{
				Type: personType, // personType has personType
			},
		},
	})
}
```

The current `AddFieldConfig()` implementation is one way to workaround that:
```go
func main() {
	// create the GraphQL object first
	personType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Person",
		Fields: graphql.FieldConfigMap{
			"name": &graphql.FieldConfig{
				Type: graphql.String,
			},
		},
	})

	// add the field config that refers to itself
	personType.AddFieldConfig("bestFriend", &graphql.FieldConfig{
		Type: personType,
	})
}
```

With this PR, it'll allow user to define cyclic field configuration like so:
```go
	personType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Person",
		Fields: (graphql.ObjectFieldMapThunk)(func() graphql.FieldConfigMap {
			return graphql.FieldConfigMap{
				"name": &graphql.FieldConfig{
					Type: graphql.String,
				},
				"bestFriend": &graphql.Field{
					return personType
				},
			}
		},
	})
```
`graphql.ObjectConfig.Fields` now is an inline function that would be resolved later at runtime.


---

We can have a discussion here to see which direction we would like to take 👍🏻
Both ways are valid and we could go either way.

I prefer the thunk/inline function way, it would be easier to write schema definition in one big chunk and it'll be much closer to how one would write it with `graphql-js`

After #68 gets merged in, probably `graphql.ObjectFieldMapThunk` can be renamed to `graphql.FieldsThunk` 
(`Thunk` seems a very JS thing though, any other suggestions?)

```go
	personType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Person",
		Fields: (graphql.FieldsThunk)(func() graphql.Fields {
			return graphql.Fields{
				"name": &graphql.Field{
					Type: graphql.String,
				},
				"bestFriend": &graphql.Field{
					return personType
				},
			}
		},
	})
```

Either way, we would need to make the way to define cyclic config consistent.

Cheers!

/cc @chris-ramon

chris-ramon:
closing this one in favor of: https://github.com/graphql-go/graphql/pull/229 — thanks a lot @atrniv :+1:

PR #76, opened 2015-11-20
Add a more detailed example

jaredonline:
I was looking for something in `./examples` that showed just about all of
the features of GraphQL, especially the Schema features. I couldn't find
it (until after I started) so I decided to implement the Star Wars
example from the Facebook Javascript GraphQL example.

I got about 2/3 of the way through this when I saw you had already done
this in your tests.

Either way, I think its a good example to include in the `./examples` folder.

### DIFF
--- a/examples/star-wars/data.json
+++ b/examples/star-wars/data.json
@@ -1,0 +1,55 @@
+{
+    "humans": {
+        "1000": {
+            "id": "1000",
+            "name": "Luke Skywalker",
+            "friends": ["1002", "1003", "2000", "2001"],
+            "appearsIn": [4, 5, 6],
+            "homePlanet": "Tatooine"
+        },
+        "1001": {
+            "id": "1001",
+            "name": "Darth Varder",
+            "friends": ["1004"],
+            "appearsIn": [4, 5, 6],
+            "homePlanet": "Tatooine"
+        },
+        "1002": {
+            "id": "1002",
+            "name": "Han Solo",
+            "friends": ["1000", "1003", "2001"],
+            "appearsIn": [4, 5, 6],
+            "homePlanet": "Corellia"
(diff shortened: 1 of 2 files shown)


chris-ramon:
Hi @jaredonline thanks a lot for your contribution! :star2: 

I agree `star-wars` is a great example! - and it would be great to have it on `./examples` directory.
I would like to propose the following:

- Use the `testutil.StarWarsSchema`, so we avoid code duplication.
- So we could end-up with something like:

`./examples/star-wars/main.go`
```go
package main

import (
	"encoding/json"
	"fmt"
	"net/http"

	"github.com/graphql-go/graphql/testutil"
)

func main() {
	http.HandleFunc("/graphql", func(w http.ResponseWriter, r *http.Request) {
		query := r.URL.Query()["query"][0]
			Schema:        testutil.StarWarsSchema,
			RequestString: query,
		})
	})

	fmt.Println("Now server is running on port 8080")
	fmt.Println("Test with Get      : curl -g 'http://localhost:8080/graphql?query={hero{name}}'")
}
```

- I removed the `Use the cannonical example from Facebook ... ` - I think is better to, not have an static url to `graphql-js` file.
- It was missing one `'` at the end of `curl -g 'http://localhost:8080/graphql?query={hero{name}}'`

chris-ramon:
@jaredonline I went ahead and created a new [PR](https://github.com/graphql-go/graphql/pull/93) in favor of this one - cherry-picked your commit and updated it based on the comments I did above, thanks again for initially working on this. :+1:

jaredonline:
Hey @chris-ramon, thanks! Sorry I went AWOL, I had a bunch of unexpected work travel right before the holidays. Thanks for doing the work to finish this up [=

PR #80, opened 2015-11-25
resolve field function can return an error

chris-ramon:
#### Details

- [x] Update `graphql.FieldResolveFn` return signature.
- [x] Update usage of `graphql.FieldResolveFn` to match new return signature.
- [x] Update `executor` to handle when any `graphql.FieldResolveFn` returns an error.
- [x] Write tests.

### DIFF
--- a/introspection.go
+++ b/introspection.go
@@ -86,24 +86,24 @@ func init() {
 		Fields: Fields{
 			"kind": &Field{
 				Type: NewNonNull(__TypeKind),
-				Resolve: func(p ResolveParams) interface{} {
+				Resolve: func(p ResolveParams) (interface{}, error) {
 					switch p.Source.(type) {
 					case *Scalar:
-						return TypeKindScalar
+						return TypeKindScalar, nil
 					case *Object:
-						return TypeKindObject
+						return TypeKindObject, nil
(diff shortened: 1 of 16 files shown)


introspection.go
```diff
@@ -86,24 +86,24 @@ func init() {
...
-						return TypeKindInputObject
+						return TypeKindInputObject, nil
 					case *List:
-						return TypeKindList
+						return TypeKindList, nil
 					case *NonNull:
-						return TypeKindNonNull
+						return TypeKindNonNull, nil
 					}
 					panic(fmt.Sprintf("Unknown kind of type: %v", p.Source))
```
Fugiman: I'd expect that when a resolver encounters an error in this new API they'd `return nil, fmt.Errorf(...)` instead of `panic(fmt.Sprintf(...))`. This would also move closer to a panic-less graphql-go.

chris-ramon: good catch! thanks @Fugiman :+1: totally agree.

Fugiman:
:+1: Looks great to me, awesome work!

chris-ramon:
> Looks great to me, awesome work!

Thanks!

PR #81, opened 2015-11-29
Add simple mutations and list example

kr15h:
This is a result of my attempt to get the mutations and lists to work. This is something that I felt seriously missing from the examples directory as with these 3 functions one can now replace a REST service almost completely. I wonder what about GraphQL and authentication...


chris-ramon:
Thanks a lot @kr15h :star2:

Neat example! :+1:

What do think if we could go ahead and rename the dir:
`./examples/mutations` -> `./examples/todo` - so it becomes the official todo example :smile: 

I think test's are failing because I merged: https://github.com/graphql-go/graphql/pull/80

kr15h:
OK @chris-cp will do.

chris-ramon:
@kr15h would you mind update it make this mergeable? so this useful example can land away. :100:

kr15h:
Will do.

kr15h:
So what about the failing tests? The `graphql.FieldResolveFn` causes some trouble. I am searching for a way to fix this, but...

chris-ramon:
@kr15h I went ahead and fixed it and did minor improvements.

So closing this in favor of [#91](https://github.com/graphql-go/graphql/pull/91). 

Thanks again! :smiley:

Issue #86, opened 2015-12-04
 introspection query fails with missing subscriptionType


bsr203:
Using introspection query gave following error
```
"Cannot query field \"subscriptionType\" on \"__Schema\"."
```

This is due to the recent upgrade to graphql with support for subscription.
https://github.com/graphql/graphql-js/pull/189/files#diff-4ac20bee447bf30479babc3b266c5145R54

we may need to handle it here
https://github.com/graphql-go/graphql/blob/master/introspection.go#L269

bsr203:
https://github.com/graphql-go/graphql/pull/87

chris-ramon:
Thanks for initially suggesting  partial `subscriptionType` support within introspection @bsr203, it's been added on #100, related conversation [here](https://github.com/graphql-go/graphql/pull/100/files#r49266565).

bsr203:
thanks chris

PR #87, opened 2015-12-04
added subscriptionType to pass introspection query

bsr203:
fixes https://github.com/graphql-go/graphql/issues/86 till we support subscription

### DIFF
--- a/introspection.go
+++ b/introspection.go
@@ -266,6 +266,14 @@ mutation operations.`,
 					return nil, nil
 				},
 			},
+			"subscriptionType": &Field{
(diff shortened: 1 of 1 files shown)


chris-ramon:
Hi @bsr203, thanks for the contribution!

If we add a `subscriptionType`, we will allow users to actually write **valid queries** that contains `subscription` keyword on it, since we did not implemented the subscription feature yet, it might be a little confusing for users I think.

bsr203:
agree, it is not a solution. but, not having this make it impossible to work with graphiql. Ideal may to return some not implemented error. Meanwhile, I may just make this change in my local version.

PR #88, opened 2015-12-04
Add json tags to correct case for FromattedError

mleonard87:
The GraphQL specification in section [7.2.2](https://facebook.github.io/graphql/#sec-Errors) says that the error response should contain entries with the keys: `message`, `locations`, `line` and `column`.

The spec also says in section [2.1.9](https://facebook.github.io/graphql/#sec-Names) the following:

> Names in GraphQL are case‐sensitive. That is to say `name`, `Name`, and `NAME` all refer to different names. Underscores are significant, which means other_name and othername are two different names.

As such, I have added tags to the relevant field declarations in the `FormattedError` type so that, when marshalled, the response will be case-correct.

### DIFF
--- a/gqlerrors/formatted.go
+++ b/gqlerrors/formatted.go
@@ -7,8 +7,8 @@ import (
 )
 
 type FormattedError struct {
-	Message   string
-	Locations []location.SourceLocation
+	Message   string                    `json:"message"`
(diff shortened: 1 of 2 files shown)


chris-ramon:
Looks good to me, thanks a lot @mleonard87! :star2:

PR #94, opened 2015-12-22
Fix hang in parser with bad tokens

samuel:
Error from `advance` were not being handled which led to an infinite
loop in the parser. The included test would hang (timing out) before the change.


chris-ramon:
Great catch @samuel! :star2:  - thanks a lot!

Looks good to me :+1:

PR #98, opened 2016-01-05
Adds Context support.

chris-ramon:
#### Details

- [x] Adds support for `Context` available on each field's `Resolve` function, eg:
  ```go
  graphq.Do({
    // ...
    Context: context.WithValue(context.Background(), "currentUser", user)
  })
  // ...
  Resolve: func(p graphql.ResolveParams) {
    u := p.Context.Value("currentUser")
  }
  ```
  - On previous related pr's, `Context` was exposed:
    - A new param of  `Resolve` function - `func(ctx context.Context, p ResolveParams)`
    - Through `graphql.ResolveParams.Info.Context`

  - I think it's more concise to expose context as field of `graphql.ResolveParams.Context` like to example showed above, if any thoughts on this please do share :+1: 
 
- Thanks a lot to @tallstreet & @augustoroman who initially proposed and worked on this on #25 and #82.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -7,6 +7,7 @@ import (
 	"regexp"
 
 	"github.com/graphql-go/graphql/language/ast"
+	"golang.org/x/net/context"
 )
 
 // These are all of the possible kinds of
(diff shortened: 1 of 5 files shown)


augustoroman:
Cool!  I like it.

- Augusto

CentaurWarchief:
Finally! :beers:

pyros2097:
Nice now I can use https://github.com/labstack/echo context instead of a rootValue

PR #96, opened 2015-12-29
Add FieldsThunk

pyros2097:

PR #100, opened 2016-01-08
Adds Fieldsthunk to ObjectConfig.

chris-ramon:
#### Details.

- [x] Updates `ObjectConfig.Fields` to be an interface, so now possibles can be:
  - `graphql.Fields` & `graphql.FieldsThunk`

- Built in top of great work from @pyros2097 on #96  :+1:

Note: 
I've removed extra code within `Object.Fields()` since `graphql.defineFieldMap()` does handle it and is being covered with:
`TestTypeSystem_ObjectsMustHaveFields_RejectsAnObjectTypeWithMissingFields`

cc/ @pyros2097

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -397,7 +397,10 @@ func doesFragmentConditionMatch(eCtx *ExecutionContext, fragment ast.Node, ttype
 		if conditionalType == ttype {
 			return true
 		}
-
+                if conditionalType.GetName() == ttype.GetName() {
+			return true
+		}
+		
 		if conditionalType, ok := conditionalType.(Abstract); ok {
 			return conditionalType.IsPossibleType(ttype)
 		}
--- a/executor.go
+++ b/executor.go
@@ -397,7 +397,7 @@ func doesFragmentConditionMatch(eCtx *ExecutionContext, fragment ast.Node, ttype
 		if conditionalType == ttype {
 			return true
 		}
-                if conditionalType.GetName() == ttype.GetName() {
+                if conditionalType.Name() == ttype.Name() {
 			return true
 		}
 		
--- a/introspection.go
+++ b/introspection.go
@@ -266,6 +266,14 @@ mutation operations.`,
 					return nil, nil
 				},
 			},
+			"subscriptionType": &Field{
+				Description: `If this server support subscription, the type that ' +
+                   'subscription operations will be rooted at.`,
+				Type: __Type,
+				Resolve: func(p ResolveParams) (interface{}, error) {
+					return nil, nil
+				},
+			},
 			"directives": &Field{
 				Description: `A list of all directives supported by this server.`,
 				Type: NewNonNull(NewList(
--- a/definition.go
+++ b/definition.go
@@ -412,10 +412,6 @@ func (gt *Object) Fields() FieldDefinitionMap {
 		configureFields = gt.typeConfig.Fields.(Fields)
 	case FieldsThunk:
 		configureFields = gt.typeConfig.Fields.(FieldsThunk)()
-	case nil:
-	default:
-		gt.err = errors.New(fmt.Sprintf("Unknown Object.Fields type: %v", reflect.TypeOf(gt.typeConfig.Fields)))
-		return nil
 	}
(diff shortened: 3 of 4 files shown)


pyros2097:
Great @chris-ramon I guess coverage has to be improved but other than that I guess we are going to reach the goal of an alpha release soon.

executor.go
```diff
@@ -397,7 +397,10 @@ func doesFragmentConditionMatch(eCtx *ExecutionContext, fragment ast.Node, ttype
 		if conditionalType == ttype {
 			return true
 		}
-
+                if conditionalType.Name() == ttype.Name() {
```
chris-ramon: @pyros2097 - I might be missing something, do you really think it should be rely only on `.Name` to say are the same types?

introspection.go
```diff
@@ -266,6 +266,14 @@ mutation operations.`,
...
 				},
 			},
+			"subscriptionType": &Field{
+				Description: `If this server support subscription, the type that ' +
+                   'subscription operations will be rooted at.`,
+				Type: __Type,
+				Resolve: func(p ResolveParams) (interface{}, error) {
+					return nil, nil
+				},
+			},
```
chris-ramon: @pyros2097 there was a [PR](https://github.com/graphql-go/graphql/pull/87) for this and I commented quote:
> If we add a subscriptionType, we will allow users to actually write valid queries that contains subscription keyword on it, since we did not implemented the subscription feature yet, it might be a little confusing for users I think.

pyros2097: I think we need to implement the subscription feature but before that we need a working graphql server. Without this the introspection query, graphiql and relay fails and we couldn't use our frontend without this. Maybe we should let the user's know that the subscription feature is not implemented but type exists in our system for compatibility's sake.

executor.go
pyros2097: I guess its not absolutely necessary but since every type has to have a unique name this should also hold valid true. It's very useful in because say we have a type ex:
`type Quiz struct {}` which I use internally I want to be able to send some of its data back to client without creating a new struct so we use a map[string]interface{} and map out the fields we want to send back but the problem arises when using relay.

When we are using nodeDefinition's it doesn't know it is a quiz type but is map[string]interface{}so for this we use the type name like this
Ofcourse we can use any other field but this is easier to reason and it adds some sort of dynamic behavoir to the strong go type struct type system.
```go
MutateAndGetPayload(params graphql.Params) (interface{}, error) {
  return map[string]interface{}{
    "__typename": "Quiz",
    "id": "quiz1",
  }
}

//and in our nodeDefinitions we resolve the type like this
NodeDefinitions = relay.NewNodeDefinitions(relay.NodeDefinitionsConfig{
	IDFetcher: func(id string, info graphql.ResolveInfo) interface{} {
		resolve := relay.FromGlobalID(id)
		switch resolve.Type {
		case "Quiz":
			return map[string]interface{}{
				"__typename": "Quiz",
				"id": "some_id",
			}
            }
    }
  },
            }
  TypeResolve: func(value interface{}, info graphql.ResolveInfo) *graphql.Object {
		switch data := value.(type) {
		case map[string]interface{}:
			switch data["__typename"] {
			case "Quiz":
				return QuizType
			}
		}
	}
})
```
This helps us in not creating to many structs for a single graphql Type.

chris-ramon: Thanks for clarification @pyros2097 - let's keep this since you've provide a handy use case.

bsr203:
https://github.com/graphql-go/graphql/pull/87 :-)
I currently do this locally to make work with graphiql.

introspection.go
chris-ramon: Alright @pyros2097 thanks for insights, I think we can keep this one as well, since there's on going graphiql support (https://github.com/graphql/graphiql/issues/55 & https://github.com/graphql/graphiql/issues/79) for graphql servers which does not support subscription feature yet.

andrewscarani:
@chris-ramon @pyros2097 Any working examples (preferably working test cases) of recursive input types using InputObjectConfigFieldMapThunk?

PR #107, opened 2016-01-22
Fix isNullish to properly handle typed nil

samuel:
The test demonstrates what the problem was. For typed nil values (e.g. fields of a struct that are pointers to other structs), isNullish wasn't properly detecting the nil value.

Also removed the NaN check for int since there's no case where it can be true as far as I know.


samuel:
Sorry for the noise, forgot this PR was open and when I merged to our fork's master it updated it.

PR #110, opened 2016-02-01
Expose DefaultResolveFn to make it wrappable outside the library.

neglectedvalue:
Hi!
I've started implementing ACL for my schema of data, and I want to be able to wrap the default resolve function for simple fields.
The wrapper is a function which checks rights for the particular field or object. If the access is not allowed, it returns an error.
Not sure if it is the right implementation. Probably, there's another way, and I would glad to hear alternatives.
Thanks!

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -397,10 +397,10 @@ func doesFragmentConditionMatch(eCtx *ExecutionContext, fragment ast.Node, ttype
 		if conditionalType == ttype {
 			return true
 		}
-                if conditionalType.Name() == ttype.Name() {
+		if conditionalType.Name() == ttype.Name() {
 			return true
 		}
-		
+
 		if conditionalType, ok := conditionalType.(Abstract); ok {
 			return conditionalType.IsPossibleType(ttype)
 		}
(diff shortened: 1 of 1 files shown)


sogko:
Hi @neglectedvalue 
The GraphQL spec does not prescribe how to deal with authorisation/authentication.

I would recommend doing it on server level (for e.g. a middleware that authenticates/authorizes an incoming request to `/graphql` before actually executing a GraphQL query).
You then can inject into the request context the user and it's relevant access / role.

This way, your GraphQL implementation just worries about how to fetch data, for a given request context.

neglectedvalue:
Hi @sogko 

Yes, this way I care only about authorization for a given request context and user.

The authentication is done by a middleware.

The thing is that the same object could have different access level for by different users, g.e. the user's email should be available only for this user and admins. Or like an object could be accessible only by its creator, so I need to compare a field like `user_id` with the authenticated user.
Currently, I'm doing it wrapping a resolve function, to check if this node and a field of graphql request are available in a given context and `graphql.FieldResolveFn.Source`. If it is not then, the wrapper will return an access denied error.

If I traversed nodes of a graphql request separately, it would be not efficient as using a wrapper because of too many SQL queries.

sogko:
Hi @neglectedvalue 

Sorry for the long wait for reply, really appreciate your contribution to the project 👍🏻

@neglectedvalue Probably it would be useful and beneficial to the project if you could provide an example to help other users relate to this use-case (there is already an `/examples/` folder that contains great stuff from other users).

@chris-ramon What do you think? It is a simple change the exposed `graphql.defaultResolveFn`. I am for this PR, especially if a working example is provided.

Cheers! 👍🏻

chris-ramon:
closing this one in favor of https://github.com/graphql-go/graphql/pull/227 — thanks a lot @neglectedvalue :+1:

PR #114, opened 2016-02-15
Add Pointer support to GraphQL

### LINKED ISSUE #101: How should pointers to standard types be handled when no `Resolve` is defined?
I know that if I had a struct like this:

```go
type User struct {
        Username string `json:"username"`
        Age      int    `json:"age"`
}
```

Then I could simply define a type with a field called `"username"` and a type of `graphql.String` and then another field `"age"` with a type of `graphql.Int` and the fields would "auto-resolve". However, I've been defining GraphQL types for AWS API return values and AWS uses a lot of pointers in their structs. For example their IDs/Names are an 'aws.String' which is basically `*string` so the pointer address is returned in the GraphQL response. Obviously this is not what is desired, but it's odd to have this all over the place:

```go
// ...
        source := params.Source.(*User)

        return *source.Username, nil
}

// ...
        source := params.Source.(*User)

        return *source.Age, nil
}
```

Do you think the standard Scalar types should support pointers as well? Or should there be a wrapper type like `Type: graphql.NewPointer(...)`?

bbuck:
This adds a new meta-type, `Pointer` to support pointer fields in structs. This allows easily converting a pattern like:

``` go
"ptr_field": &graphql.Field{
        Type: graphql.String,
                source := params.Source.(*SomeType)

                return *source.PtrField
        },
},
```

To simply:

``` go
"ptr_field": &graphql.Field{
        Type: graphql.NewPointer(graphql.String),
},
```

I realize that there is no spec for pointers but pointers aren't part of every language. I think it's important that we allow support for this due to significant amounts of boilerplate for simple pointer fields. I don't think there is really a better way to represent this outside of creating this new type, but I feel like it's an elegant solution.

This PR would close #101

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -164,6 +164,7 @@ var _ Named = (*Interface)(nil)
 var _ Named = (*Union)(nil)
 var _ Named = (*Enum)(nil)
 var _ Named = (*InputObject)(nil)
+var _ Named = (*Pointer)(nil)
 
 func GetNamed(ttype Type) Named {
 	unmodifiedType := ttype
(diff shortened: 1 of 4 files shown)


sogko:
Hi @bbuck 

Thanks for the PR, really appreciate the effort you contributed to the project.

Currently I have reservations about introducing a new type to `graphql-go` that goes beyond the current published spec. GraphQL is still very much in its early infancy and things does get change pretty quickly (we barely have enough time to achieve parity with `graphql-js` at the moment lol).

I think currently #101 fits the use case for a __user-defined Scalar type__. This means that the user can create a new Scalar type named Pointer through existing API `graphql.NewScalar()`, similar how built-in types like Float, String and Int are defined in the library.
(Refer to: https://github.com/graphql-go/graphql/blob/master/scalars.go)

If anyone need a complete example of how to implement a custom scalar type using `graphql.NewScalar()`, I can spend some time writing that a gist for it.

Thanks again!

zbintliff:
This was superceded by https://github.com/graphql-go/graphql/pull/199

This was done without introducing a new point, but accepting pointers or scalars and coercing them correctly.

bbuck:
This issue needs to be closed, you are correct. The intent here was to do this explicitly, instructing those viewing your Schema that a pointer is expected to be returned. I'm not opposed to the solution presented in #199 though, as I just wanted the problem to be resolved!

PR #121, opened 2016-04-13
Fix default values of input objects.

svenschneider:
This pull request fixes an issue where the default value of an input object's field was ignored. The issue is triggered under the following conditions:
1. An object's field contains arguments which are input objects.
2. The input objects contain fields for which a default value is specified.
3. The query does not provide an input object's field for which a default value is specified.
In this situation, the input object's field should be set to the default value, but was not.

Any comments on this pull request are welcome.

Best regards
Sven Schneider

### DIFF
--- a/executor_test.go
+++ b/executor_test.go
@@ -1230,6 +1230,56 @@ func TestQuery_ExecutionDoesNotAddErrorsFromFieldResolveFn(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


chris-ramon:
closing this one in favor of https://github.com/graphql-go/graphql/pull/226 — thanks a lot @svenschneider :+1:

PR #137, opened 2016-06-10
Improve lexer performance by using a byte array as source

Matthias247:
I did some peformance testing for graphql-go and found out that it was superslow for my use-case (about 300ms for processing a simple query). After using the go profiling tools it was obvious that the Garbage Collector takes most of the time for it (> 250ms).

I have then viewed through the graphql related code and have seen that the Lexer produces an extreme amount of garbage. The reason is that the Lexer stores the request as a `string` type but converts it to ´[]rune` every time it wants to read a char. This will cause an allocation and copy for reading each character!

I have now made changes were the `Source` is directly storing a `[]byte` array, which can be indexed without performance penalty and which is never converted to a rune array. I changed some functions to read runes from the given offset and the source array and to be able to skip over them even if they are multibyte values.

This pull request improves performance for me by factor 30! This means I'm coming to 10ms for processing a request, which is still not stellar but more workable. In the new performance traces the lexer is no longer the most time consuming part.

The code was not heavily reviewed, so I suggest someone else also takes a look at it before merging. But tests are working and my application with it is working too.

### DIFF
--- a/gqlerrors/syntax.go
+++ b/gqlerrors/syntax.go
@@ -4,10 +4,11 @@ import (
 	"fmt"
 	"regexp"
 
+	"strings"
+
 	"github.com/graphql-go/graphql/language/ast"
 	"github.com/graphql-go/graphql/language/location"
 	"github.com/graphql-go/graphql/language/source"
-	"strings"
 )
 
 func NewSyntaxError(s *source.Source, position int, description string) *Error {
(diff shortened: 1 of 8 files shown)


sogko:
Hi @Matthias247 

Thanks for taking your time to dive deep into this and submitting a PR, appreciate it very much 👍

Briefly glancing through the changes, it looks good. I'll take the time to review it ASAP.

Cheers!

PR #147, opened 2016-07-18
Fix nil panic in UniqueOperationNamesRule with two anonymous operations

titanous:
`node.Name` is nil when the operation is anonymous. Use the parent AST node instead in this case.

Found with go-fuzz.

### DIFF
--- a/rules.go
+++ b/rules.go
@@ -2,13 +2,14 @@ package graphql
 
 import (
 	"fmt"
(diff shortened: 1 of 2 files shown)


chris-ramon:
closing this one in favor of https://github.com/graphql-go/graphql/pull/224 — thanks a lot @titanous

PR #154, opened 2016-07-24
Experimental/Need feedback: Implement pluggable batching/dedup/concurrent fetch like facebook/dataloader

bigdrum:
I have been considering how to implement the idea of facebook/dataloader with graphql in go (which address issues like #106, #132), and this PR shows the progress so far.

This PR is mainly a demonstration on how it could work. And so far I'm happy with the idea.

This requires some minimum changes to the graphql package (just to inject a executor with a simple interface). The real concurrency feature is implemented in the godataloader library that I wrote separately, which can be plugged into graphql.

examples/dataloader contains an actual example, which demonstrate the following features:
- Opt-in.
- Concurrently loading multiple data.
- Load duplicate requests once.
- Batching multiple requests.
- No unnecessary goroutines are spawn unless concurrency actually happen.
- Synchronous programming style (without Promise-like object).

The test might best demonstrate the behavior.

``` go
func TestQuery(t *testing.T) {
    schema := dataloaderexample.CreateSchema()
    r := dataloaderexample.RunQuery(`{
        p1_0: post(id: "1") { id author { name }}
        p1_1: post(id: "1") { id author { name }}
        p1_2: post(id: "1") { id author { name }}
        p1_3: post(id: "1") { id author { name }}
        p1_4: post(id: "1") { id author { name }}
        p1_5: post(id: "1") { id author { name }}
        p2_1: post(id: "2") { id author { name }}
        p2_2: post(id: "2") { id author { name }}
        p2_3: post(id: "2") { id author { name }}
        p3_1: post(id: "3") { id author { name }}
        p3_2: post(id: "3") { id author { name }}
        p3_3: post(id: "3") { id author { name }}
        u1_1: user(id: "1") { name }
        u1_2: user(id: "1") { name }
        u1_3: user(id: "1") { name }
        u2_1: user(id: "3") { name }
        u2_2: user(id: "3") { name }
        u2_3: user(id: "3") { name }
    }`, schema)
    if len(r.Errors) != 0 {
        t.Error(r.Errors)
    }
    t.Error(r)
    // The above query would produce log like this:
    // 2016/07/23 23:49:31 Load post 3
    // 2016/07/23 23:49:31 Load post 1
    // 2016/07/23 23:49:31 Load post 2
    // 2016/07/23 23:49:32 Batch load users [3 1 2]
    // Notice the first level post loading is done concurrently without duplicate.
    // The user loading is also done in the same fashion, but batched fetch is used instead.
    // TODO: Make test actually verify the logged behavior.
}
```

This PR is not meant to be merged yet (at least the example probably doesn't fit to be part of this project because of its external dependency, unless the godataloader library is moved).

The godataloader library is probably a little bit more complicated than one would expect. And I'm happy to explain it if anyone finds this idea interesting.

### DIFF
--- a/examples/dataloader/example.go
+++ b/examples/dataloader/example.go
@@ -43,6 +43,10 @@ var userDB = map[string]*user{
 		ID:   "2",
 		Name: "John",
 	},
+	"3": &user{
+		ID:   "3",
+		Name: "Kate",
+	},
 }
 
 var loaderKey = struct{}{}
@@ -62,14 +66,18 @@ func newLoader(sch *dataloader.Scheduler) *loader {
 			id := key.(string)
 			return dataloader.NewValue(postDB[id], nil)
 		})),
-		userLoader: dataloader.New(sch, dataloader.Parallel(func(key interface{}) dataloader.Value {
-			// In practice, we will make remote request (e.g. SQL) to fetch post.
+		userLoader: dataloader.New(sch, func(keys []interface{}) []dataloader.Value {
+			// In practice, we will make remote request (e.g. SQL) to fetch multiple users.
 			// Here we just fake it.
-			log.Print("Load user ", key)
+			log.Print("Batch load users ", keys)
 			time.Sleep(time.Second)
-			id := key.(string)
-			return dataloader.NewValue(userDB[id], nil)
-		})),
+			var ret []dataloader.Value
+			for _, key := range keys {
+				id := key.(string)
+				ret = append(ret, dataloader.NewValue(userDB[id], nil))
+			}
+			return ret
+		}),
 	}
 }
 
@@ -144,6 +152,22 @@ func CreateSchema() graphql.Schema {
 					}
 					dl := getDataLoader(p.Context)
 					return dl.postLoader.Load(id).Unbox()
+				},
+			},
+			"user": &graphql.Field{
+				Type: userType,
+				Args: graphql.FieldConfigArgument{
+					"id": &graphql.ArgumentConfig{
+						Type: graphql.String,
+					},
+				},
+				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
+					id, ok := p.Args["id"].(string)
+					if !ok {
+						return nil, nil
+					}
+					dl := getDataLoader(p.Context)
+					return dl.userLoader.Load(id).Unbox()
 				},
 			},
 		}})
--- a/examples/dataloader/example.go
+++ b/examples/dataloader/example.go
@@ -1,0 +1,194 @@
+package dataloaderexample
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"golang.org/x/net/context"
+
+	"github.com/bigdrum/godataloader"
+	"github.com/graphql-go/graphql"
+)
+
+var postDB = map[string]*post{
+	"1": &post{
+		ID:       "1",
+		Content:  "Hello 1",
+		AuthorID: "1",
+	},
+	"2": &post{
+		ID:       "2",
+		Content:  "Hello 2",
+		AuthorID: "1",
+	},
+	"3": &post{
+		ID:       "3",
+		Content:  "Hello 3",
+		AuthorID: "2",
+	},
+	"4": &post{
+		ID:       "4",
+		Content:  "Hello 4",
+		AuthorID: "2",
+	},
+}
+
+var userDB = map[string]*user{
+	"1": &user{
+		ID:   "1",
+		Name: "Mike",
+	},
+	"2": &user{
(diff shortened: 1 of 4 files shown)


bsr203:
It may be a while till I expiriment with it, but like how you done it all with minimal changes and also github.com/bigdrum/godataloader is kinda small. thanks for improving this library. cheers.

teloon:
This look really interesting. Anything we're missing here?

bigdrum:
FYI, I've been using this at a small scale production environment. And I haven't seen any bugs so far. The only problem that I still consider this as experiment is that for a object with only trivial resolve function (one that without expensive IO), the dataloader would still kick off a goroutine (one per parent object, not one per item, though, so not that many). We don't see any performance issue so far but it would be nice to provide some hint to optimize that away.

ajackson-cpi:
Any status on this? I'm interested in parallel processing for my use-cases. GoRoutines are quite light if they're short-lived, so I wouldn't spend too much time on them.

tonyghita:
Have you looked at https://github.com/nicksrandall/dataloader? This library should probably be orthogonal to any dataloader

bigdrum:
https://github.com/nicksrandall/dataloader only dispatches the batch when batch size reaches some limit or some pre-configured amount of wait time has passed. This will introduce unnecessary delay. On the other hand, facebook's dataloader.js and my implementation do not have such issue.

tonyghita:
@bigdrum this is because facebook's implementation takes advantage of the javascript event loop.  Golang (to my understanding) doesn't have such an event loop, so @nicksrandall's implementation approximates this with the batch size/time elapsed mechanism (both parameters are configurable to your use case).

bigdrum:
Right my implementation of data loader implements a custom scheduler to
achieve the similar effect.

bigdrum:
To be more elaborated, in essence, there are two kinds of tasks, the tasks that collect what need to be loaded, and the tasks of the fetching. If we can prioritize the collection task over the fetching task, we can batch as much as possible.

FB's dataloader.js takes advantage of js env where a closure can be schedule at the end of the event loop, which essentially allow them to schedule the lower priority fetch tasks after the collection task. So event loop is not essence here, the essence is the ability to dispatch tasks in a certain order.

Go doesn't support custom go routine scheduling, if it does, we could create a schedule domain, dispatch go routine with different strict priority. So I ended up implement a custom scheduling mechanism that allows running different closure with a custom order, which has two queues internally, to allow executing the tasks with two tier priority. Thus to achieve the same result.

@nicksrandall's implementation achieve this by manually delay the fetch task with a fixed amount of time, to allow the collection tasks to be executed before that. But since it doesn't know when the collection is done, it pays the cost of waiting blindly.

It is a trade-off, @nicksrandall is definitely much easier to understand, and my implementation is harder, but to me, the precise scheduling without compromising the latency is exactly what I really want and inspired from FB's dataloader.js. Without which, I might just use node.js for graphql implementation. I'm presenting a way to achieve the same effect even though the underlying runtime is so different between node.js and go.

In the end, what I'm proposing here is just a interface to allow such different execution model to be injected. People can choose different actual executor as they see fits their needs.

(Sorry for potential grammar errors as I'm not a English native speaker.)

ajackson-cpi:
I'm not sure where to look, but this scenario sounds like it needs a sync.WaitGroup{} that sees all the .Add() calls before the collector tasks start, and something does .Wait() before starting the fetchers.  The collector tasks can call .Done()

That's the scatter-gather pattern in GoLang.

nicksrandall:
@bigdrum I like your approach of creating a custom scheduler. I know this a first draft but a few features that I think are critical for a datoader are max batch size and max time out. I worry that your implementation under load could grow unbounded. That feature should be pretty easy to add.

bigdrum:
@nicksrandall Right that's very good suggestion, and my approach does introduce more footprint comparing to yours as it creates new goroutines when it sees the need to yield to another graphql tree to collect more tasks (but note at a given moment only one goroutine is active, all others are waiting to be scheduled). 

The interesting thing is that this logic could be put into the custom scheduler. For example, we could provide some fairness to the low priority queue, like if the high priority queue size reaches some limit, dispatch the task in low priority queue. This allows the pending batch to be flushed. Same for the time based criteria.

Since the dataloader/scheduler is per graphql query. We could also introduce a shared semaphore state so that the total pending tasks of a server could not exceed some given limit.

ajackson-cpi:
I'm interested in seeing this merged too. What's the status?

sandeepone:
Any Update? 👍

divoxx:
Any updates on this?

shengwu:
Any updates on this vs parallel resolution?

chris-ramon:

PR #168, opened 2016-11-13
Add 1.7 and tip to build matrix for Go

Tonkpils:
Let's start keeping up to date with the latest Go versions. Travis will at least complain if there's something wrong in a later version that we can fix.


chris-ramon:
LGTM, let's move forward to keep up-to-date with last go versions, thanks @Tonkpils

Issue #184, opened 2017-02-08
Context Should Respect Deadline


RaniSputnik:
When passing a context into the schema with a deadline, the deadline of the context should cancel resolutions that are yet to be started. `graphql.Do` should complete on-time with a timeout error reported if there were still fields to be resolved.

```
// Query type includes a field that won't resolve within the 5 second deadline
var queryType = graphql.NewObject(
	graphql.ObjectConfig{
		Name: "Query",
		Fields: graphql.Fields{
			"hello": &graphql.Field{
				Type: graphql.String,
				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
                                        time.Sleep(10 * time.Second)
					return "world", nil
				},
			},
		},
	})
schema, _ := graphql.NewSchema(graphql.SchemaConfig{
	Query: queryType,
})

ctx := context.WithTimeout(context.Background(), 5*time.Second)
result := graphql.Do(graphql.Params{
	Schema:        schema,
	RequestString: "{hello}",
	Context: ctx,
})
```

PR #185, opened 2017-02-08
Respect Context Deadline

RaniSputnik:
Have added context timeout to graphql.Execute, this seemed the most appropriate place to put it. A couple of concerns I have;

* Nested resolve functions could still be started after a context has already been cancelled
* Not sure of the performance implications of these changes - doesn't look too bad to me but I hasn't been benchmarked
* The go func inside of Execute's pretty messy now, the same select is repeated three times
* The tests run slower after this change as they need to wait for the context timeout. I've kept this to 100ms for now, so not a massive deal - but it might be better to kick off this test in parallel.

Feedback very much appreciated!

Closes https://github.com/graphql-go/graphql/issues/184

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -31,7 +31,6 @@ func Execute(p ExecuteParams) (result *Result) {
 	}
 
 	resultChannel := make(chan *Result)
-	doneChannel := make(chan struct{})
 
 	go func(out chan<- *Result, done <-chan struct{}) {
 		result := &Result{}
@@ -81,7 +80,7 @@ func Execute(p ExecuteParams) (result *Result) {
 		case <-done:
 		}
 
-	}(resultChannel, doneChannel)
+	}(resultChannel, ctx.Done())
 
 	select {
 	case <-ctx.Done():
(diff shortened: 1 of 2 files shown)


chris-ramon:
> Have added context timeout to graphql.Execute, this seemed the most appropriate place to put it. A couple of concerns I have;
> 
> Nested resolve functions could still be started after a context has already been cancelled
> Not sure of the performance implications of these changes - doesn't look too bad to me but I hasn't been benchmarked
> The go func inside of Execute's pretty messy now, the same select is repeated three times
> The tests run slower after this change as they need to wait for the context timeout. I've kept this to 100ms for now, so not a massive deal - but it might be better to kick off this test in parallel.
> Feedback very much appreciated!

Thanks a lot @RaniSputnik!

> Nested resolve functions could still be started after a context has already been cancelled
I think it's fine, if graphql results contain errors, clients will either retry or something else.

> Not sure of the performance implications of these changes - doesn't look too bad to me but I hasn't been benchmarked
I don't think there are perf. implications, if so we will handle it on new PR's

> The go func inside of Execute's pretty messy now, the same select is repeated three times
Could not agree more, looking forward to code clean-up inside the `Execute` function.

> The tests run slower after this change as they need to wait for the context timeout. I've kept this to 100ms for now, so not a massive deal - but it might be better to kick off this test in parallel.
IMHO 100ms is fine for now.

LGTM :+1: :ship:

PR #197, opened 2017-04-08
Add BindFields function tool

ahmadmuzakki29:
I've been working with graphql-go lately and realize I'm pretty lazy to create object type for every struct I had. So I made a binder/wrapper function to easily bind all fields in my struct into graphql object.

So if you have this struct 
```go
type Person struct {
    Name    string  `graph:"name"`
    Address string  `graph:"address"`
}
```
you can create person type with 
```go
personType := graphql.NewObject(graphql.ObjectConfig{
        Name:   "Person",
        Fields: BindFields(Person{}),
    })
```

in file `util_test.go` I made an example with Person struct that looked like this:
```go
type Person struct {
	Human //embedded Fields
	Name    string   `json:"name"` // standard scalar type
	Home    Address  `json:"home"` // struct type
	Hobbies []string `json:"hobbies"` // slice of scalar type 
	Friends []Friend `json:"friends"` // slice of struct type
}

type Human struct {
	Alive  bool    `json:"alive"`
	Age    int     `json:"age"`
	Weight float64 `json:"weight"`
}

type Friend struct {
	Name    string `json:"name"`
	Address string `json:"address"`
}

type Address struct {
	Street string `json:"street"`
	City   string `json:"city"`
}
```

because this function will scan through empty struct so this function cannot accept recursive type. that would make a stack-overflow error. And one more utility function which is `BindArgs` that can bind the args depending on provided params.

### DIFF
--- a/util.go
+++ b/util.go
@@ -76,21 +76,22 @@ func getGraphType(tipe reflect.Type) Output {
 }
 
 func getGraphList(tipe reflect.Type) *List {
-	switch tipe {
-	case reflect.TypeOf([]int{}):
-	case reflect.TypeOf([]int8{}):
-	case reflect.TypeOf([]int32{}):
-	case reflect.TypeOf([]int64{}):
-		return NewList(Int)
-	case reflect.TypeOf([]bool{}):
-		return NewList(Boolean)
-	case reflect.TypeOf([]float32{}):
-	case reflect.TypeOf([]float64{}):
-		return NewList(Float)
-	case reflect.TypeOf([]string{}):
-		return NewList(String)
+	if tipe.Kind() == reflect.Slice {
+		switch tipe.Elem().Kind() {
+		case reflect.Int:
+		case reflect.Int8:
+		case reflect.Int32:
+		case reflect.Int64:
+			return NewList(Int)
+		case reflect.Bool:
+			return NewList(Boolean)
+		case reflect.Float32:
+		case reflect.Float64:
+			return NewList(Float)
+		case reflect.String:
+			return NewList(String)
+		}
 	}
-
 	// finaly bind object
 	t := reflect.New(tipe.Elem())
 	name := strings.Replace(fmt.Sprint(tipe.Elem()), ".", "_", -1)
--- a/util.go
+++ b/util.go
@@ -3,8 +3,8 @@ package graphql
 import (
 	"encoding/json"
 	"fmt"
-	"github.com/graphql-go/graphql"
 	"reflect"
+	"strings"
 )
 
 const TAG = "json"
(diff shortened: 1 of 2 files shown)


util.go
```diff
@@ -0,0 +1,159 @@
...
+		return Boolean
+	case reflect.Slice:
+		return getGraphList(tipe)
+	}
+	return String
+}
+
+func getGraphList(tipe reflect.Type) *List {
+	switch tipe {
+	case reflect.TypeOf([]int{}):
```
bbuck: Not a fan of all the `reflect` usage here, you've already got a `reflect.Type`, you can use it. like here `case tipe.Kind() == reflect.Slice && tipe.Elem().Kind() == reflect.Int` and other tests `case tipe.Kind() == reflect.Slice && tipe.Elem().Kind() == reflect.Int8`, etc...

bbuck:
I'm still not a fan of auto-binding nested structs, which is exactly the opposite of how I (and all the demos I've ever seen of how to use/configure GraphQL go. The point is to make sub records searchable in their own right. For example, you _should_ be able to bind recursively. If you have a `me` query endpoint that returns a `Person` type, and a `Person struct` has a field `Friends []Person` that should return a `[Person]` and GraphQL's system handle it perfectly. That is what it's designed to do (not just recursive, but nested/linked data in this manner).

ahmadmuzakki29:
if you mean nested structs by `Person` struct is `Human` then `Human` struct is supposed to be on the same level. so the `Person` struct is like having all the fields that `Human` has.
```go
type Person struct{
	Alive  bool    `json:"alive"`
	Age    int     `json:"age"`
	Weight float64 `json:"weight"`
	Name string    `json:"name"`
       Address Address `json:"address"`
       // and so on ....
}
```
in that struct the sub record is address, so yes you can query `Address` in their own right. 
But if you take example like `Friends` with type of `Person` this function will loop forever because it will bind the `Person` recursively.
appreciated your input, but I'm interested to hear another people opinion about this approach.

zeucxb:
Waiting

ahmadmuzakki29:
@zeucxb I updated the PR and hope it can be accepted to merge, if it's not feel free to just copy util.go and place it in your package

chris-ramon:
This util function is very awesome!, @ahmadmuzakki29 thanks a lot for taking the time.

I believe this is the initial function for more useful always optional time-savers functions, let's improve the recursive part and other improvements on new PR's!, thanks! :ship:

ahmadmuzakki29:
Sure thanks!

axetroy:
I got the error

```bash
"reflect: call of reflect.Value.NumField on ptr Value"
```

and here is my struct, can anyone help?

```go
type Post struct {
	ID        string       `json:"id"`
	UID       string       `json:"uid"`
	Cid       string       `json:"cid"`
	ShortId   string       `json:"shortId"`
	Title     string       `json:"title"`
	Summary   string       `json:"summary"`
	Content   string       `json:"content"`
	Gallery   []string     `json:"gallery"`
	Like      int32        `json:"like"`
	Comment   int32        `json:"comment"`
	Share     int32        `json:"share"`
	View      int32        `json:"view"`
	Tags      []string     `json:"tags"`
	IsShow    bool         `json:"isShow"`
	IsActive  bool         `json:"isActive"`
	IsTop     bool         `json:"isTop"`
	CreatedAt rpcbase.Time `json:"createdAt"`
	UpdatedAt rpcbase.Time `json:"updatedAt"`
}
```

bbuck:
@axetroy How are you calling the tool? Also it might be better to create your own issue for this, as this issue is related to merging the tool in, not bugs and/or issues arising from using it.

ahmadmuzakki:
@axetroy I suspect rpcbase.Time causing it. please create the fields manually for CreatedAt and UpdatedAt, and set it like this
```
CreatedAt rpcbase.Time `json:"-"`
UpdatedAt rpcbase.Time `json:"-"`
```

limoli:
Same problem with a struct having custom fields `time.Time`.
I think that bindFields **doesn't handle** time types.

PR #199, opened 2017-04-30
fixed issue with pointer to scalar values

jhaynie:
If you have a response struct like the following:

```
struct Foo {
   Bar *string  `json:"bar"`
}
```

And `Bar` is `nil`. The result will be incorrectly encoded on returning.  The result will be `\u003cnil\u003e` instead of `null` (when serialized to JSON).

This fix will fix `isNullish` to check for `*string` type and other primitive types.

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -25,52 +25,78 @@ func coerceInt(value interface{}) interface{} {
 			return nil
 		}
 		return value
+	case *int:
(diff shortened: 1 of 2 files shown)


chris-ramon:
LGTM :+1: — thanks a lot @jhaynie

PR #203, opened 2017-05-25
Add checking if node != nil - NewError, gqlerrors

mwarzynski:
Following query causes panic: `{graphql_is:great(sort:ByPopularity)}{stars}`.
I added checking if node is nil, before executing method on it.

Also, I am not sure if error message `There can only be one operation named ""` is okay (empty operation name). But definitely it's better than panic.


gqlerrors/error.go
```diff
@@ -38,6 +42,9 @@ func NewError(message string, nodes []ast.Node, stack string, source *source.Sou
 	}
 	if len(positions) == 0 && len(nodes) > 0 {
 		for _, node := range nodes {
+			if node == nil || reflect.ValueOf(node).IsNil() {
```
mwarzynski: For me, it's cleaner to separate checking if node is nil from checking if an attribute (or something that GetLoc() returns) is nil. Also, three conditions within the single line is quite too much.

mwarzynski: @bbuck Do you want me to put these three conditions within the single line?

bbuck: If you're asking because you think it's holding up the merge, it's not. I was just posing the question.

chris-ramon:
lgtm :+1: thanks a lot @mwarzynski — closing this one in favor of https://github.com/graphql-go/graphql/pull/221

Issue #205, opened 2017-06-03
Using pointers with Enum types results in nils


threadwaste:
This is an extension of the issue highlighted by #114, and subsequently unintentionally fixed in #199. Specifically, the use of a pointer in a response type results in no resolution when e.g. serializing an Enum type.

Example, an Enum type of:

```go
var enum = graphql.NewEnum(graphql.EnumConfig{
    Name: "MyEnum",
    Values: graphql.EnumValueConfigMap{
        "first-enum": &graphql.EnumValueConfig{Value: 1}
    }
}
````

...used with a response struct as follows:

```go
type Response struct {
    MyEnum *int
}
```

theRemix:
so does #220 and #206 close this issue?

PR #206, opened 2017-06-03
Support use of pointers with Enum types

threadwaste:
This allows a response struct to used pointer types with Enum field types as shown in #205. In the case of `Serialize`, we simply leverage `reflect.Indirect` in the face of what is essentially a generic for which we don't need the exact type. `ParseValue` thankfully just directly works with strings, so we us a straight type switch.


chris-ramon:
closing this one in favor of: https://github.com/graphql-go/graphql/pull/220

PR #209, opened 2017-06-16
Add DateTime scalar type

jeromer:
I needed a `DateTime` scalar type. So here it is. I tester the serialization part, but did not test the deserialization do you have any unit test I should update in that regard ?

Thanks :)


bbuck:
Nice to see the work done, for sure. Unfortunately I don't think we'll get to see this merged because it's not an actual type in GraphQL (per the spec), and this repository is striving to keep with the spec. (I know because I attempted to support pointers via a new "scalar" for explicit reasons but never saw the work merged).

Thanks for the effort though! If possible you can package up your code in a gist and share it that way. I'm sure you're not the only person out there who wants/needs a date/time scalar!

jeromer:
> Nice to see the work done, for sure. Unfortunately I don't think we'll get to see this merged because it's not an actual type in GraphQL (per the spec), and this repository is striving to keep with the spec. (I know because I attempted to support pointers via a new "scalar" for explicit reasons but never saw the work merged).

Ok.

> Thanks for the effort though! If possible you can package up your code in a gist and share it that way. I'm sure you're not the only person out there who wants/needs a date/time scalar!

Here  you go https://gist.github.com/jeromer/881f1403abdfaa2dfad3fd270a4b4692

:)

chris-ramon:
I believe this and other types are always used on most of the projects while building schemas.

I think we should include this within the [util.go](https://github.com/graphql-go/graphql/blob/master/util.go) file.

chris-ramon:
closing this one in favor of: https://github.com/graphql-go/graphql/pull/219

Issue #212, opened 2017-07-10
replace x/net/context with context package


mbertschler:
Thanks for this great library, I really like to use it!
At some point this library should use the `context` standard library package instead of the old `golang.org/x/net/context` one. The new package was introduced in Go 1.7, and is now available for almost a year already.

This is a little problem for me since I use the standard context package in my code, and the old x/net package then causes import conflicts, which requires me to change the import to something like netcontext so that I can use both at the same time.

If you decide that this is something we should do, then I could send a PR that changes this.

bbuck:
Make a PR for it!

chris-ramon:
closed via: https://github.com/graphql-go/graphql/pull/214 thx a lot!

PR #214, opened 2017-07-16
Use standard library's "context" package

exklamationmark:
While going through the source code, I noticed that it's still using `golang.org/x/net/context`.

This could be swapped with `golang.org/pkg/context` without any further code changes. However, this  mean the package would need Go 1.7+ to work.

wdyt?

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -1,12 +1,12 @@
 package graphql
 
 import (
+	"context"
 	"fmt"
 	"reflect"
 	"regexp"
 
 	"github.com/graphql-go/graphql/language/ast"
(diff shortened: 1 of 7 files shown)


exklamationmark:
This should address https://github.com/graphql-go/graphql/issues/212

chris-ramon:
lgtm :+1: 

- we are running travis tests against go1.7 as lower supported Go version: https://github.com/graphql-go/graphql/blob/master/.travis.yml#L4
- also if just created a release for < go1.7 users: https://github.com/graphql-go/graphql/releases/tag/v0.7.0

PR #216, opened 2017-07-18
Examples: use '.Get("query")' instead of '["query"][0]'

WindomZ:
In examples, 
if the query(`RawQuery`) values does not contain the key `"query"`:
1. Use `r.URL.Query()["query"][0]`, throws runtime error: `index out of range`.
1. Use `r.URL.Query().Get("query")`, gets the `graphql.Result.Errors` string such as `[Must provide an operation.]`. Maybe more friendly.

It works in go1.5.


chris-ramon:
lgtm, good catch!, thanks a lot @WindomZ 👍

PR #217, opened 2017-07-23
no dependencies

chris-ramon:
#### Overview
- vendor: removes vendor directory, until now we were using two dependencies:
  - `https://github.com/kr/pretty` - for debugging, to check diff from expect & actual tests, we can always use this awesome lib, when working on this lib for debugging.
  - `golang.org/x/net/context"` - for context support, but dependency was removed via: https://github.com/graphql-go/graphql/pull/214, since go1.7+ included in the std lib.
- testutil: updated Diff, to just return want & got, instead of the actual diff, now output looks like this, Eg:
```
--- FAIL: TestBasicGraphQLExample (0.00s)
        graphql_test.go:131: wrong result, query: { hello }, graphql result diff: [
                got: &{map[hello:world] []}
                want: map[hello:worldz]
                ]
```

#### Test plan
- [x] `go test ./...`


chris-ramon:
removed in favor of: https://github.com/graphql-go/graphql/pull/218

PR #229, opened 2017-08-06
support for cyclic fields on defintion

chris-ramon:
#### Overview
- Built on top of https://github.com/graphql-go/graphql/pull/73 — thanks a lot @atrniv :+1: 
- Adds support for cyclic fields.

#### Test plan
- [x] `go test ./...`

### DIFF
--- a/definition_test.go
+++ b/definition_test.go
@@ -619,3 +619,25 @@ func TestTypeSystem_DefinitionExample_IncludesFieldsThunk(t *testing.T) {
 		t.Fatalf("Unexpected result, Diff: %v", testutil.Diff(fieldMap["s"].Type, someObject))
 	}
 }
+
+func TestTypeSystem_DefinitionExampe_AllowsCyclicFieldTypes(t *testing.T) {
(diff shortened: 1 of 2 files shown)


definition_test.go
```diff
@@ -619,3 +619,25 @@ func TestTypeSystem_DefinitionExample_IncludesFieldsThunk(t *testing.T) {
...
+func TestTypeSystem_DefinitionExampe_AllowsCyclicFieldTypes(t *testing.T) {
+	personType := graphql.NewObject(graphql.ObjectConfig{
+		Name: "Person",
+		Fields: (graphql.FieldsThunk)(func() graphql.Fields {
+			return graphql.Fields{
+				"name": &graphql.Field{
+					Type: graphql.String,
+				},
+				"bestFriend": &graphql.Field{
+					Type: personType,
```
stupschwartz: @chris-ramon `Type: personType` is using the `personType` defined in `union_interface_test.go`, not the local variable defined just above.  If `personType` hadn't been defined in that other file, you would get an `undefined: personType` error.

PR #238, opened 2017-08-17
Add optional Type to FormattedError

maxsz:
This is how github's API exposes machine-readable error strings. We need this in our API to better be able to handle specific errors in the client. This is just a proposal to start a discussion on how to best handle this requirement 😃 

For example:

Query:
```graphql
{
  repository(owner:"bla",name:"blubb") {
    id
  }
}
```

Response:
```json
{
  "data": {
    "repository": null
  },
  "errors": [
    {
      "message": "Could not resolve to a Repository with the name 'blubb'.",
      "type": "NOT_FOUND",
      "path": [
        "repository"
      ],
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

### DIFF
--- a/gqlerrors/error.go
+++ b/gqlerrors/error.go
@@ -17,7 +17,12 @@ type Error struct {
 	Positions     []int
 	Locations     []location.SourceLocation
 	OriginalError error
+	Extensions    ErrorExtensions
 }
+
+// ErrorExtensions contains custom extensions that are passed through to the
+// client.
+type ErrorExtensions map[string]interface{}
 
 // implements Golang's built-in `error` interface
 func (g Error) Error() string {
--- a/gqlerrors/formatted.go
+++ b/gqlerrors/formatted.go
@@ -1,6 +1,7 @@
 package gqlerrors
 
 import (
+	"encoding/json"
 	"errors"
 
 	"github.com/graphql-go/graphql/language/location"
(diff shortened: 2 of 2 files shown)


maxsz:
What do you think @sogko, @chris-ramon?

christianpv:
@maxsz I believe we have to do something like this instead: https://github.com/graphql/graphql-js/commit/0b2e81c768cd4da96082cec07a973b2017dcbe97#diff-f8e4bedf89ec2148ccb1602bfc6a7ae7

maxsz:
@christianpv looks good! I've updated the PR 👍

dvic:
I went with a different approach in #279 and chose just to store the original error with `FormattedError` and move the responsibility of rendering errors differently up the chain. This enables me easily to render my domain errors differently, since I now have access to the actual underlying error. Any thoughts on this approach?

chris-ramon:
Thanks a lot @maxsz! — latest spec of GraphQL details how to expose additional information about errors via `extensions`:
```javascript
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018"
      }
    }
  ]
}
```

We have a PR for it, so closing this one in favor of: https://github.com/graphql-go/graphql/pull/363

PR #240, opened 2017-08-22
possible fix

agoalofalife:
I sent
```
{
  human(id: "1000") {
    name
  }
}
```
And get
```
{
  "data": {
    "human": null
  },
  "errors": [
    {
      "message": "interface conversion: interface {} is string, not int",
      "locations": []
    }
  ]
}
```
don't know whether it was intended, but I fixed

### DIFF
--- a/testutil/testutil.go
+++ b/testutil/testutil.go
@@ -300,7 +300,10 @@ func init() {
 					},
 				},
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
-					id, _ := strconv.Atoi(p.Args["id"].(string))
+					id, err := strconv.Atoi(p.Args["id"].(string))
+					if err != nil {
+					  return nil, err
+					}
 					return GetHuman(id), nil
 				},
 			},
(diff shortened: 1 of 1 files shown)


testutil/testutil.go
```diff
@@ -300,7 +300,8 @@ func init() {
 					},
 				},
 				Resolve: func(p graphql.ResolveParams) (interface{}, error) {
-					return GetHuman(p.Args["id"].(int)), nil
+					id, _ := strconv.Atoi(p.Args["id"].(string))
```
adelowo: Since `Resolve`'s signature already returns an error, why suppress the error returned from `strconv.Atoi` ?

chris-ramon: Thanks a lot for going ahead and fixing it @agoalofalife — good call @adelowo, could you update it to:
```go
id, err := strconv.Atoi(p.Args["id"].(string))
if err != nil {
  return nil, err
}
return GetHuman(id), nil
```

agoalofalife: @chris-ramon I changed

chris-ramon:
lgtm :+1: — thanks a lot @agoalofalife

PR #241, opened 2017-08-24
Cleaner default resolver, cache FieldDefMap Get, preallocate

ajackson-cpi:
Returning 26k rows for me was taking 8 seconds of post-Resolve() processing. I reduced that to 305ms by:

- Returning the cached FieldDefMap (7 seconds)
- Preallocating result map/array structures (5ms or so)
- DefaultResolver logically identical, but far more efficient: uses reflect.FieldByName instead of checking in a loop. Allocates fewer variables.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -441,6 +441,9 @@ func (gt *Object) String() string {
 	return gt.PrivateName
 }
 func (gt *Object) Fields() FieldDefinitionMap {
+	if len(gt.fields) != 0 {
+		return gt.fields
+	}
 	var configureFields Fields
 	switch gt.typeConfig.Fields.(type) {
 	case Fields:
--- a/executor.go
+++ b/executor.go
@@ -827,12 +827,17 @@ func defaultResolveTypeFn(p ResolveTypeParams, abstractType Abstract) *Object {
 func DefaultResolveFn(p ResolveParams) (interface{}, error) {
 	// try to resolve p.Source as a struct first
 	sourceVal := reflect.ValueOf(p.Source)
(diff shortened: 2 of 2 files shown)


bbuck:
It might be worth building a lazy cache for different types as well, so we don't have to keep looping fields all the time. Maybe not as part of this PR (nice work) but just a general remark.

edpark11:
Thanks for writing this! An (admittedly large) graphql list query was taking 10s.
This piece in particular (returning the cached field definition map) resulted in a 8x speedup. Possible to accept that piece of the patch? The default resolver patch currently breaks against the master branch but the cached field definition (just 3 lines) works fine.
```
func (gt *Object) Fields() FieldDefinitionMap {                                      
  if len(gt.fields) != 0 {
    return gt.fields
  }
```

ajackson-cpi:
It looks like the latest master has brought equivalent changes. My performance evaluation shows it quite similar.

chris-ramon:
Thank you for given latest changes on master a try @ajackson-cpi! there was a similar optimization introduced by https://github.com/graphql-go/graphql/pull/265, a couple of opt. missing from this PR:
- https://github.com/graphql-go/graphql/pull/241/commits/7efae5ecbaada0757ab923d3ea66ca19907f8956#diff-58bccd1b2cfaf06d4d4ac893e4712c88L281
- https://github.com/graphql-go/graphql/pull/241/commits/7efae5ecbaada0757ab923d3ea66ca19907f8956#diff-58bccd1b2cfaf06d4d4ac893e4712c88L793

We might want to add those via a new PR.

PR #249, opened 2017-10-03
travis: update go versions

dvrkps:


chris-ramon:
awesome!, so we can run tests against latest versions! — thanks a lot @dvrkps :+1:

PR #253, opened 2017-10-22
executor: adds panic handler

chris-ramon:
#### Overview
- graphql: adds `PanicHandler` type.
- graphql: adds `Params.PanicHandler` field.
- executor: calls PanicHandler when there's a panic while resolving a field.

#### Test plan
- Unit test.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -22,7 +22,6 @@ type ExecuteParams struct {
 	// information to resolve functions.
 	Context context.Context
 
-	// PanicHandler will be called if any of the resolvers or mutations panic
 	PanicHandler func(ctx context.Context, err interface{})
 }
 
--- a/executor.go
+++ b/executor.go
@@ -22,7 +22,7 @@ type ExecuteParams struct {
 	// information to resolve functions.
 	Context context.Context
 
-	PanicHandler func(ctx context.Context, err interface{})
+	PanicHandler PanicHandler
 }
 
 func Execute(p ExecuteParams) (result *Result) {
--- a/executor.go
+++ b/executor.go
@@ -589,8 +589,7 @@ func resolveField(eCtx *ExecutionContext, parentType *Object, source interface{}
 	})
 
 	if resolveFnError != nil {
-		eCtx.Errors = append(eCtx.Errors, gqlerrors.FormatError(resolveFnError))
-		return result, resultState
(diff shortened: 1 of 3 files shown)


scottjg:
@chris-ramon thanks for your help with this patch!

i was going over your changes and i have a question... originally my patch would only call the panic handler on an actual panic. it appears with your changes, it gets called when a resolver returns any errors (maybe due to reverting the change here https://github.com/graphql-go/graphql/pull/253/commits/b63c4f02154a8075d775df2a8ccf960485717907#diff-58bccd1b2cfaf06d4d4ac893e4712c88L592 ). was that intentional? if so, is it possible to distinguish between a panic and an error in the panic handler? i think everything gets flattened to a graphql formatted error type so i'm not sure if it's possible.

ahmadmuzakki:
I stuck around on this problem also, I need to know where the stacktrace of the panic in my resolver. is this patch possible to solve that?
I used to manually add
```
defer func() {
	if r := recover(); r != nil {
		log.Println("panic: ", string(debug.Stack()))
		panic(r)
	}
}()
```
on every resolver.
So with this handler I should print with debug.Stack() in the handler?
and also why this PR still not merged to master yet, or should I use another version for using this patch?

rodrigo-brito:
@chris-ramon  Do you have any forecast for the release?

Niondir:
looking forward to get this merged.

executor.go
```diff
@@ -512,7 +518,9 @@ func resolveField(eCtx *ExecutionContext, parentType *Object, source interface{}
 	var returnType Output
 	defer func() (interface{}, resolveFieldResultState) {
 		if r := recover(); r != nil {
-
+			if eCtx.PanicHandler != nil {
+				eCtx.PanicHandler(eCtx.Context, r)
```
rodrigo-brito: It will capture all erros from GraphQL. If you see [here](https://github.com/graphql-go/graphql/blob/451533e82374cae4032ce83553bb374af94c5a62/executor.go#L536). All errors was given by panic(). Changing the `return` to value (`error`), can solve the problem.

phifty:
any updates on this?

jpascal:
any updates on this?

maddsua:
Looking forward to have this merged, really need an option to add custom panic handler 👀

PR #261, opened 2017-12-12
Add support for block strings to language package

jamesdphillips:
## Reference material

- Accepted RFC https://github.com/facebook/graphql/pull/327
- Documentation http://facebook.github.io/graphql/draft/#sec-String-Value
- Block String Value http://facebook.github.io/graphql/draft/#BlockStringValue()
- Ex. Javascript implementation https://github.com/graphql/graphql-js/blob/v0.11.7/src/language/lexer.js

### DIFF
--- a/language/lexer/lexer.go
+++ b/language/lexer/lexer.go
@@ -3,6 +3,8 @@ package lexer
 import (
 	"bytes"
 	"fmt"
+	"regexp"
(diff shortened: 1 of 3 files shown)


chris-ramon:
Awesome!, thanks a lot @jamesdphillips :ship:

PR #264, opened 2017-12-18
Fix utility function

ahmadmuzakki:
Fix json tag with multiple value (comma separated). The graphql name will return error if the json tag value containing comma
add more case for graph integer type


chris-ramon:
LGTM :+1: thanks a lot @ahmadmuzakki

PR #266, opened 2018-01-05
Hide effectively private structs

fd:


chris-ramon:
LGTM :+1: — thanks a lot @fd :ship:

PR #274, opened 2018-01-30
Support for anonymous contained structs

Oscil8:
These should behave similarly to the equivalent case in
https://golang.org/pkg/encoding/json/


chris-ramon:
Thanks a lot @Oscil8! — closing this one in favor of: https://github.com/graphql-go/graphql/pull/371

PR #277, opened 2018-02-05
Include stack trace in errors

gaelduplessix:
Hey,

Currently, if a runtime error occurs during the execution of a query, the only information available looks like this:
```json
{
  "data": null,
  "errors": [
    {
      "message": "runtime error: invalid memory address or nil pointer dereference",
      "locations": []
    }
  ]
}
```

Which is quite tricky to debug.

This PR adds a "StackTrace" field to the `FormattedError` object. This field is not exposed to the JSON version, so it doesn't expose possibly sensible debug information but can be used for logs.
Currently the field is only populated in one place but I guess it makes sense to populate it wherever there is a `recover()` call.
Because the call to `debug.Stack()` may be expensive, maybe we should add an option to enable/disable it ? Not sure how we should expose this.

Do you think this feature make sense ?

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"runtime/debug"
 	"strings"
 
 	"github.com/graphql-go/graphql/gqlerrors"
@@ -61,7 +62,9 @@ func Execute(p ExecuteParams) (result *Result) {
 				if r, ok := r.(error); ok {
 					err = gqlerrors.FormatError(r)
 				}
-				exeContext.Errors = append(exeContext.Errors, gqlerrors.FormatError(err))
+				formattedErr := gqlerrors.FormatError(err)
+				formattedErr.StackTrace = string(debug.Stack())
+				exeContext.Errors = append(exeContext.Errors, formattedErr)
 				result.Errors = exeContext.Errors
 				select {
 				case out <- result:
--- a/gqlerrors/formatted.go
+++ b/gqlerrors/formatted.go
@@ -7,8 +7,9 @@ import (
 )
 
 type FormattedError struct {
-	Message   string                    `json:"message"`
-	Locations []location.SourceLocation `json:"locations"`
+	Message    string                    `json:"message"`
+	Locations  []location.SourceLocation `json:"locations"`
+	StackTrace string                    `json:"-"`
 }
 
 func (g FormattedError) Error() string {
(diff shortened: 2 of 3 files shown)


dvic:
I  just went with the approach of letting it implement the `Causer` interface: https://github.com/qdentity/graphql-go/blob/master/gqlerrors/formatted.go

In this way, you can extract your own errors thrown in the resolvers and render them appropriately.

gaelduplessix:
@dvic Indeed, sounds more generic.

Still, I think the trace should be included anyway. In my case, I had a runtime exception happening inside the lib (this is probably a bug btw, if performing a mutation query when none is defined in the schema)

So to include the stack trace, we would build a custom error type that returns the trace in the Cause() method ? Or the user could retrieve the underlying error type and call `Trace()` on it directly ?

dvic:
@gaelduplessix I modified that code to use `github.com/pkg/errors` instead of the standard library `errors` package. The package `github.com/pkg/errors` stores the stack trace for you (you can see it when you print the error with `%v` or `%+v`).

dvic:
By the way, https://github.com/qdentity/graphql-go also contains changes w.r.t. how resolvers are called. I took the changes from #213 and added the concurrent resolving of list values #132. This is a perfect match with https://github.com/nicksrandall/dataloader. There were some data races but I managed to fix these as well, all tests are passing with the -race flag included. If anybody is interested in this I can open a PR.

gaelduplessix:
@dvic Oh, sounds neat :) Well, the error improvement seems quite an isolated issue. Would it make sense to extract it to another PR ?

dvic:
:) Those fixes are only useful/needed if #213 or #132 are merged, however, it is not clear at the moment which path will be chosen. So it's best to wait until a decision has been made before merging those fixes into the appropriate branch.

I did however submit a PR (#278) just now that makes sure the tests check for race conditions, so the issue is caught as soon as concurrency is added to the project.

gaelduplessix:
Well, improvements on errors are useful on their own, wouldn't it make sense to have these changes isolated from the rest (adding concurrency to the resolver) so they can be merged sooner ?

Is there any ETA on when #213 or #132 are going to be merged ? (I'm quite new to this project so I don't really have any context on how the development is going)

dvic:
I have no idea on the ETA of either #213 #132, but I'll cook up a PR for the changes w.r.t. the errors (it's basically [this](https://github.com/qdentity/graphql-go/commit/055b8d1daad5dd0f313a6634635e02ca39c9d212) commit).

dvic:
Created a PR #279 for the changes of embedding errors in `FormattedError`.

dvic:
@gaelduplessix did you have time to check out #279? Will it work for your use case?

PR #278, opened 2018-02-05
Pass the '-race' flag to goveralls when running tests on TravisCI

dvic:


nicolaiskogheim:
Why?

nicolaiskogheim:
OK, I guess you wanted to catch problems with https://github.com/graphql-go/graphql/pull/132 and https://github.com/sogko/graphql/pull/26
Seems valid to me.

dvic:
Yes, all future PR's that add concurrency will benefit from this.

PR #288, opened 2018-03-01
Implement resolving fields from FieldResolver source values

dvic:
This allows users to return types that can implement their own `FieldResolveFn`.


chris-ramon:
awesome!, def. agree on having an optional `FieldResolver`, very go idiomatic.
thanks a lot @dvic :ship:

Issue #156, opened 2016-08-03
Modifying Context during a resolve function to transfer data down a tree


bbuck:
If, for example, you had some query like:

```
first {
  second {
    third {
      a
      b
      c
    }
  }
}
```

And during the resolve function for `first` you want to carry over some data for `third` to use during resolution you might think you could do something like:

```go
params.Context = context.WithValue(params.Context, someContextKey, someValue)
```

But of course this won't work at all. I think it might be worth finding a way to allow something like this to happen, otherwise it would seem that `third` has to always be built in such a way that it can't know anything about the tree it's being requested from other than through arguments (and if you're having to repeat the same argument per tier that might become problematic).

I'm not sure if the graphql-js "root value" solution allows this to happen (and if it doesn't, then I suppose there is no reason for this project too) it just seems useful.

What are your thoughts about this?

axetroy:
graphql-js support add some property on context. so you can transfer data down a tree.

Can you found a way for this?

axetroy:
I found the way to transfer data down a tree, but It is too tedious.

Query With this options

```go
		result := graphql.Do(graphql.Params{
			Schema:         ql.Schema,
			RequestString:  graphQuery.Query,
			Context:        context,
			VariableValues: graphQuery.Variables,
			OperationName:  graphQuery.OperationName,
			RootObject: map[string]interface{}{
				"context":  context,
			},
		})

```

The parent resolve
```go
			Resolve: func(p graphql.ResolveParams) (interface{}, error) {

				type GraphContext struct {
					context interface{}
				}

				rootValue := p.Info.RootValue.(map[string]interface{})
				ctx := rootValue["context"].(*gin.Context)

				ctx.Request = ctx.Request.WithContext(context.WithValue(ctx.Request.Context(), "testKey", "testValue"))

				var (
					token = p.Args["token"]
				)
				_, _, err := service.ParseToken((token).(string), service.TagNameUser)

				if err != nil {
					return nil, service.Err_Invalid_Token
				}

				return meFields, nil
			},
```

The child resolve

```go
			Resolve: func(p graphql.ResolveParams) (interface{}, error) {

				rootValue := p.Info.RootValue.(map[string]interface{})
				ctx := rootValue["context"].(*gin.Context)

				c := ctx.Request.Context()

				fmt.Println(c.Value("testKey"))
			},
```

lukaspj:
Is this thread-safe?
[...]

PR #292, opened 2018-03-12
Keep original error in FormattedError

ruben-podadera:
Hi all, 

First, thanks for this lib. The work you've done is amazing.

The purpose of this PR is to keep the original error in the `FormattedError` struct so we can use it later.

Why would I do this : lets say I've got books. Some of them are public, other none. As I only get the book id in the resolver, I can not do the public/private check before calling `results := graphql.Do(...)`. But once in the resolver I can not set the response code (and I should not, its not its responsability). Thus, after the call, I check on `results.Errors`. And with my PR I can do something like that :
```golang
    Schema:        schema,
    RequestString: r.URL.Query().Get("query"),
    Context:       c,
})

if len(result.Errors) > 0 {
    authorisationError, isOk := result.Errors[0].OriginalError.(AuthorisationError)
    if isOk {
        w.WriteHeader(401)
    }
    else {
        w.WriteHeader(400)
    }
    fmt.Printf("wrong result, unexpected errors: %v", result.Errors)
}

```

The other way would be to pass it through the RootObject (as mentionned [here](https://github.com/graphql-go/graphql/issues/156)), but I think that its a better thing to keep an information we already have instead of destroying it and use and other mechanism to pass it.

What do you think ?

Ruben

### DIFF
--- a/gqlerrors/formatted.go
+++ b/gqlerrors/formatted.go
@@ -7,8 +7,9 @@ import (
 )
 
 type FormattedError struct {
-	Message   string                    `json:"message"`
-	Locations []location.SourceLocation `json:"locations"`
+	Message       string                    `json:"message"`
+	Locations     []location.SourceLocation `json:"locations"`
+	OriginalError error                     `json:"-"`
 }
 
 func (g FormattedError) Error() string {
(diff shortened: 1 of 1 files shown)


ruben-podadera:
Very weird bug, this PR makes all the test fail. And I dont have the time to fix them

PR #302, opened 2018-03-24
allow resolvers as pointer methods

### LINKED ISSUE #301: Resolvers can't be implemented with pointer methods
I'm trying to make my structs self describing even though they have non-trivial fields and they need resolvers, but I noticed that I HAVE to make the resolver a non-pointer method for it to work. Example:
```go
type Thing struct{name string}
// resolver triggered:
func (t thing) Resolve(p graphql.ResolveParams) (interface{}, error)  {
    return t.name, nil
}
// resolver not triggered:
func (t *thing) Resolve(p graphql.ResolveParams) (interface{}, error)  {
    return t.name, nil
}
```
Sorry for the limited code example, but I think you can see what I mean. It's clear when you look at the code that the pointer is dereferenced before the interface is checked: https://github.com/graphql-go/graphql/blob/4536a44b434ef76ee6693f9ad1daf2ead3fd2c91/executor.go#L845

gracenoah:
fix #301

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -832,18 +832,18 @@ type FieldResolver interface {
 // and returns it as the result, or if it's a function, returns the result
 // of calling that function.
 func DefaultResolveFn(p ResolveParams) (interface{}, error) {
-	// try to resolve p.Source as a struct first
 	sourceVal := reflect.ValueOf(p.Source)
+	// Check if value implements 'Resolver' interface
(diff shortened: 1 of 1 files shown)


chris-ramon:
Thanks a lot @gracenoah!, also this is backwards compatible with non-pointer resolvers, we have a test for it: https://github.com/graphql-go/graphql/blob/master/executor_test.go#L1694

PR #305, opened 2018-03-24
DefaultResolveFn: Use case insensitive comparison

abihf:
To be able to use struct type for resolver without using field tags

schema:
```graphql
type User {
  id: ID!
  name: String
}
```

Go:
```go
type User struct {
  ID string
  Name string
}


chris-ramon:
LGTM :+1: — thanks a lot @abihf

PR #311, opened 2018-04-08
Add pointer support to BindFields

blitzprog:
This adds support for pointers in the supplied struct to the BindFields utility function.


chris-ramon:
LGTM :+1: — thanks a lot @blitzprog

PR #324, opened 2018-05-16
SDL - Separate multiple inherited interfaces with `&`

jamesdphillips:
Previously interfaces were implemented with:

```graphql
type Foo implements Bar, Baz { field: Type }
```

Replaced by:

```graphql
type Foo implements Bar & Baz { field: Type }
```

--- 
Closely mirrors reference implementation: https://github.com/graphql/graphql-js/pull/1169.


chris-ramon:
LGTM :+1: — thanks a lot @jamesdphillips!

Issue #323, opened 2018-05-10
Arrays are not recognized as iterable


alexflint:
In my code I use the following type to represent points in 2d
```
type Vec2 [2]float64
```
I realize that graphql does not have any explicit representation for fixed-size arrays so I'm happy to expose this in the api as simple a list of floats, ie
```
graphql.NewList(graphql.Float)
```
However, when I register this field, I get an error saying that my field is not iterable. If instead I return a slice from my resolver then I get the expected behavior. This is rather frustrating as I have to manually register a resolver everywhere that I return a `Vec2` to turn the array into a slice.

I think this library should recognize fixed-size arrays as iterable.

alexflint:
Here is the line that causes the problem:

https://github.com/graphql-go/graphql/blob/master/executor.go#L785

PR #327, opened 2018-05-25
treat arrays as iterable

alexflint:
This pr changes the executor to treat Go arrays as iterable. This means that if a user defines a schema including a `graphql.List`, and the value for that list is an array, then the array will be serialized as a json list.

Solves #323


glaslos:
LGTM

alexflint:
Ping @glaslos

PR #332, opened 2018-06-11
Adds an AddFieldConfig method to InputObjects

robdaemon:
Thunks don't always work for adding self-referential fields
to an InputObject.

Example:

```
inputObj := graphql.NewInputObject(graphql.InputObjectConfig{
    Name: "MyObject",
    Fields: (graphql.InputObjectConfigFieldMapThunk)(func() graphql.InputObjectConfigFieldMap {
        return graphql.InputObjectConfigFieldMap{
            "parent": &graphql.InputObjectFieldConfig{
                Type: inputObj,
            },
        }
    }),
})
```

Go does not this definition, but you can work around this with AddFieldConfig.

This makes an InputObject work like other Object types.


chris-ramon:
LGTM :+1:  — thanks a lot @robdaemon

PR #335, opened 2018-06-16
gofmt

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

https://blog.golang.org/go-fmt-your-code

---
> I made this PR with a project going on over at https://github.com/rotblauer/gofmt-att, and it's definitely a work in progress. So if I got something wrong, or this is annoying at all, please file an issue over there and we'll sort it out.


chris-ramon:
Thanks a lot @whilei :+1:

PR #341, opened 2018-06-28
Fixes #340. Thunk results are completed as per non-thunk results.

### LINKED ISSUE #340: Thunks for object types are not being resolved.
If an object has a field that is itself an object type, and it has a resolve function that returns a thunk, the result value of the thunk is returned in the result without itself being resolved.

To say another way, if a resolve function returns a thunk (`func() interface{}`), then whatever is returned by the thunk is just marshalled into JSON and included in the result.

Consider the following code fragment:

```go
	barType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Bar",
		Fields: graphql.Fields{
			"bazA": &graphql.Field{
				Type: graphql.String,
			},
		},
	})

	fooType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Foo",
		Fields: graphql.Fields{
			"bar": &graphql.Field{
				Type: barType,
					var bar struct {
						BazA string
					}
					bar.BazA = "A"

					thunk := func() interface{} { return &bar }
					return thunk, nil
				},
			},
		},
	})

	queryType := graphql.NewObject(graphql.ObjectConfig{
		Name: "Query",
		Fields: graphql.Fields{
			"foo": &graphql.Field{
				Type: fooType,
					var foo struct{}
					return foo, nil
				},
			},
		},
	})

	schema, _ := graphql.NewSchema(graphql.SchemaConfig{
		Query: queryType,
	})
	expectNoError(err)

	query := "{ foo { bar { bazA } } }"
		Schema:        schema,
		RequestString: query,
	})
```

You would expect that the result would be:

```json
{
    "foo": {
        "bar": {
            "bazA": "A"
        }
    }
}
```

Where in fact the result looks like:

```json
{
    "foo": {
        "bar": {
            "BazA": "A"
        }
    }
}
```
Note the "BazA" instead of the "bazA". The result of the thunk has just been marshalled as JSON and not interpreted as a graphql object.

jjeffery:
Note that the following tests previously passed and have been modified.

* TestLists_NullableListOfNonNullArrayOfFunc_ContainsNulls
* TestLists_NonNullListOfNonNullArrayOfFunc_ContainsNulls

It seems to me that both of these tests should have failed previously, as they involved passing nil values where nil values were not permitted. Please advise if my understanding is incorrect.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -612,12 +612,14 @@ func completeValueCatchingError(eCtx *executionContext, returnType Type, fieldAS
 func completeValue(eCtx *executionContext, returnType Type, fieldASTs []*ast.Field, info ResolveInfo, result interface{}) interface{} {
 
 	resultVal := reflect.ValueOf(result)
-	if resultVal.IsValid() && resultVal.Type().Kind() == reflect.Func {
+	for resultVal.IsValid() && resultVal.Type().Kind() == reflect.Func {
 		if propertyFn, ok := result.(func() interface{}); ok {
-			return propertyFn()
+			result = propertyFn()
+			resultVal = reflect.ValueOf(result)
+		} else {
+			err := gqlerrors.NewFormattedError("Error resolving func. Expected `func() interface{}` signature")
+			panic(gqlerrors.FormatError(err))
 		}
-		err := gqlerrors.NewFormattedError("Error resolving func. Expected `func() interface{}` signature")
-		panic(gqlerrors.FormatError(err))
 	}
 
 	// If field type is NonNull, complete for inner type, and throw field error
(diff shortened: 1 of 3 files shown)


jjeffery:
Reading the GraphQL spec (June 2018), section [3.12.1](http://facebook.github.io/graphql/June2018/#sec-Combining-List-and-Non-Null). 
The following entry applies to the `TestLists_NullableListOfNonNullArrayOfFunc_ContainsNulls` test:

[Int!] | [1, 2, null] | null (With logged coercion error)
-- | -- | --

This PR changes the test to return the result consistent with the GraphQL spec.

When it comes to the test `TestLists_NonNullListOfNonNullArrayOfFunc_ContainsNulls`, the relevant entry in section 3.12.1 is:

[Int!]! | [1, 2, null] | Error: Item cannot be null
-- | -- | --

This one might not be quite as compliant, but it is closer than the test prior to this PR.

Please advise if anyone has a different understanding.

chris-ramon:
LGTM :+1: — thanks a lot @jjeffery! 
closes: https://github.com/graphql-go/graphql/issues/340

PR #342, opened 2018-06-28
Improve test coverage for coercing scalars.

jjeffery:
Added tests for coerceInt, coerceFloat and coerceBool. 

Found some missing case statements during testing.

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -21,6 +21,8 @@ func coerceInt(value interface{}) interface{} {
 			return 1
 		}
 		return 0
+	case *bool:
+		return coerceInt(*value)
 	case int:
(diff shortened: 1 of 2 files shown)


scalars.go
```diff
@@ -153,6 +191,9 @@ func coerceFloat(value interface{}) interface{} {
 	case *string:
 		return coerceFloat(*value)
 	}
+
+	// TODO: check the graphql spec: coerceInt returns nil if it
```
chris-ramon: - Yes, we should return `nil`.
- Addressed by :https://github.com/graphql-go/graphql/pull/352

chris-ramon:
LGTM :+1: — thanks a lot @jjeffery !

PR #357, opened 2018-07-18
Another asynchronous resolution PR. But using JS-like promises instead of goroutines.

ccbrown:
Like everyone else, I'm in need of batching, deduplication, and concurrency in my query execution.

But the other pull requests all look like they utilize goroutine-based approaches. This pull request does not use goroutines. Instead, it works a bit more like the JavaScript reference implementation using a notion of promises. Here's how it works in terms of usage:

Resolvers that can execute asynchronously return a channel. To be explicit, a `graphql.ResolvePromise`. So you might have a loader that creates and returns a new `graphql.ResolvePromise` like so:

```go
{
    id, _ := p.Args["id"].(string)
    loader := p.Context.Value("loader").(*Loader)
    return loader.LoadNodeByGlobalId(id)
  },
}
```

```go
func (l *Loader) LoadNodeByGlobalId(id string) (graphql.ResolvePromise, error) {
  promise := make(graphql.ResolvePromise, 1)
  ...
  l.pendingNodes[id].Promises = append(l.pendingNodes[id].Promises, promise)
  return promise, nil
}
```

How does the loader fulfill the promise without a goroutine? When the query execution gets "stuck" and there are no more resolvers that it can invoke, it invokes an "idle handler" to let it know that one or more pending promises need to be fulfilled before it can continue. So your params can now specify a scheduler for your loader like so:

```go
p.IdleHandler = loader.Schedule
```

When the handler is invoked, it expects one or more promises to be fulfilled. Afterwards, it'll resume resolving values until it gets stuck again.

So your loader implementation might at a high level do something like this:

* If one or more pending node is cached, fulfill its promises, then return.
* If one or more pending node is for whatever reason not batchable, fulfill its promises, then return.
* Fulfill promises for a batch of nodes, then return.

And so in this way, you can really control when and how your resolutions happen. And there are no goroutines involved or multithreading concerns. Of course, you can use goroutines if you want to though.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -592,6 +592,16 @@ type ResolveInfo struct {
 	VariableValues map[string]interface{}
 }
 
+type ResolveResult struct {
+	Value interface{}
+	Error error
+}
+
+// When returned from a resolve function, ResolvePromise indicates that the resolution will be done
+// asynchronously. When used, an IdleHandler should be specified. This handler must fulfill one or
+// more promises each time it is invoked.
+type ResolvePromise chan *ResolveResult
+
 type Fields map[string]*Field
 
 type Field struct {
--- a/executor.go
+++ b/executor.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/graphql-go/graphql/gqlerrors"
 	"github.com/graphql-go/graphql/language/ast"
+	"github.com/graphql-go/graphql/promise"
 )
 
 type ExecuteParams struct {
@@ -17,6 +18,8 @@ type ExecuteParams struct {
 	AST           *ast.Document
 	OperationName string
 	Args          map[string]interface{}
+
+	IdleHandler func()
 
 	// Context may be provided to pass application-specific per-request
 	// information to resolve functions.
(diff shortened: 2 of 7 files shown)


ccbrown:
@chris-ramon 

> I'm thinking that might not be a simple API for graphql-go users, since Go does not have Promises natively, therefore Go users are not familiar with it

It might feel more at home with Go users if we just used the term "channel" instead of "promise". I don't think this sort of channel usage should feel foreign at all to anyone interested in concurrency.

> I'm thinking that something like a previous PR proposed, would match your use cases somehow ?

There are two downsides to that approach that I really want to avoid:

* It requires a new goroutine to be created for each field that returns a function. On a busy server, this is significantly more expensive than creating a channel.
* It requires batching to be done with either a timeout or a pending node threshold. There's no way for a dataloader to know precisely when no more nodes will be requested of it, so the batching algorithm can never be optimal.

chris-ramon:
Thanks a lot @ccbrown! :+1:  — Closing this one in favor of https://github.com/graphql-go/graphql/pull/388

PR #368, opened 2018-07-24
examples: adds crud example

maxchagin:


examples/crud/main.go
```diff
@@ -0,0 +1,225 @@
...
+				},
+			},
+			Resolve: func(params graphql.ResolveParams) (interface{}, error) {
+				id, _ := params.Args["id"].(int)
+				product := Product{}
+				for i, p := range products {
+					if int64(id) == p.ID {
+						product = products[i]
+						// Remove from product list
+						products = append(products[:i], products[i+1:]...)
```
chris-ramon: Here is better to use something like [`Filtering without allocating`](https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating) when removing items within a loop, just as general recommendation.

```go
productsUpdated := products[:0]
for _, p := range products {
  if int64(id) != p.ID {
    productsUpdated = append(productsUpdated, p)
  }
}
```

Issue #365, opened 2018-07-23
"Runtime error: Nil pointer reference" while serializing null string pointer


dreamer-89:
Do we need to add null pointer handling in graphql/scalers.go ?

func coerceString(value interface{}) interface{} {
	if v, ok := value.(*string); ok {
		return *v
	}
	return fmt.Sprintf("%v", value)
}

chris-ramon:
Thanks a lot for filing the issue @dreamer-89 — this was addressed via: https://github.com/graphql-go/graphql/pull/369, so closing the issue.

PR #369, opened 2018-07-26
Scalar nil pointer safety

sfriedel:
This fixes a regression when trying to serialize nil pointer values in scalars reported in #365.

The regression was introduced by the merge of #363 because previously the panic in the coercion functions was ignored in `completeValueCatchingError` and set the field to nil/null.

The change introduces explicit handling for the nil pointers in the scalar coercion functions to prevent the panic from happening.

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -22,6 +22,9 @@ func coerceInt(value interface{}) interface{} {
 		}
 		return 0
 	case *bool:
+		if value == nil {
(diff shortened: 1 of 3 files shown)


chris-ramon:
Thanks lot @sfriedel! this is definitely a great improvement, now it won't bite us again since you added tests for it. :shield: :ship:

PR #370, opened 2018-07-27
Enum nil pointer safety

sfriedel:
Similar to #369 this fixes another issue where Enum values would panic when trying to be resolved on a nil pointer.

I've also added tests for the Union, Interface and List types for their nil pointer handling but they did not need any fixes.


chris-ramon:
Awesome!, thanks a lot @sfriedel :+1:

PR #373, opened 2018-07-28
replace regexp.Compile with regexp.MustCompile

Quasilyte:
Since error is ignored, it's better to use MustCompile.

Found using https://go-critic.github.io/overview#regexpMust-ref


chris-ramon:

PR #374, opened 2018-07-28
replace len(x)<=v with len(x)==v

Quasilyte:
Length can't be negative, so use more appropriate operator.

Found using https://go-critic.github.io/overview#sloppyLen-ref


chris-ramon:

Quasilyte:
@chris-ramon thanks for warm comments. ^^
Really makes these contributing sessions enjoyable.

PR #375, opened 2018-07-28
simplify single case type switches

Quasilyte:
Found such switches using https://go-critic.github.io/overview#singleCaseSwitch-ref


chris-ramon:

PR #377, opened 2018-08-01
Fixed #355 handle null pointer lists

### LINKED ISSUE #355: null lists of non primitive objects are not handled
The following unit test demonstrates the problem:

    func TestLists_NullableListOfInt_ReturnsNull(t *testing.T) {
        ttype := graphql.NewList(graphql.Int)
        type dataType *[]int

        var data dataType

        expected := &graphql.Result{
            Data: map[string]interface{}{
                "nest": map[string]interface{}{
                    "test": nil,
                },
            },
        }
        checkList(t, ttype, data, expected)
    }

Expected functionality is that "test" is nil, instead, an error is generated:

		got: [User Error: expected iterable, but did not find one for field DataType.test.]
		want: []

The problem is that ```isNullish()``` doesn't look at the interface pointer types.  Potentially you should do a:

```
if !value.IsValid() {
   return false
} 
```

before the switch statement.

koblas:


koblas:
@chris-ramon Any chance of looking at this as well?

PR #379, opened 2018-08-11
Added Custom Error Formatter

racerxdl:
I added support for a custom error formatter. Sometimes you come with custom `error` objects that might need special parsing to add correct `extensions` in a generic way by object type.

This implementation doesn't change anything to the flow if no Custom Handler is specified since it uses by default the `FormatError` method. However if you specify, it will run your custom function expecting a `FormattedError` to return.

Just a quick example:

```go
func CustomFormat(err error) gqlerrors.FormattedError {
	switch err := err.(type) {
	case *gqlerrors.Error:
                // My custom error object returned in the mutation / query is encapsulated as OriginalError
		cf := CustomFormat(err.OriginalError)
		cf.Locations = err.Locations
		cf.Path = err.Path
		return cf
	case *QuantoError.ErrorObject:
                // Here I can call my custom ErrorObject ToFormattedError that will return an FormattedError with proper extensions set.
		return err.ToFormattedError()
	default:
		return gqlerrors.FormatError(err)
	}
}
```

Then you can use something like: 
```
	// execute graphql query
	params := graphql.Params{
		Schema:         		*h.Schema,
		RequestString:  		opts.Query,
		VariableValues: 		opts.Variables,
		OperationName:  		opts.OperationName,
		Context:        		ctx,
		CustomErrorFomatter: 	CustomFormat,
	}
	result := graphql.Do(params)
```
And everything will work.

I'm also opening a PR in graphql-go/handler since it also benefits in this custom handler. I will put a example there as well.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -12,12 +12,12 @@ import (
 )
 
 type ExecuteParams struct {
-	Schema        Schema
-	Root          interface{}
-	AST           *ast.Document
-	OperationName string
-	Args          map[string]interface{}
-
+	Schema               Schema
+	Root                 interface{}
+	AST                  *ast.Document
+	OperationName        string
+	Args                 map[string]interface{}
+	CustomErrorFormatter func(err error) gqlerrors.FormattedError
 	// Context may be provided to pass application-specific per-request
 	// information to resolve functions.
 	Context context.Context
@@ -30,13 +30,19 @@ func Execute(p ExecuteParams) (result *Result) {
 		ctx = context.Background()
 	}
 
+	var errFormatter = gqlerrors.FormatError
+
+	if p.CustomErrorFormatter != nil {
+		errFormatter = p.CustomErrorFormatter
+	}
+
 	resultChannel := make(chan *Result)
 	result = &Result{}
 
 	go func(out chan<- *Result, done <-chan struct{}) {
 		defer func() {
 			if err := recover(); err != nil {
-				result.Errors = append(result.Errors, gqlerrors.FormatError(err.(error)))
+				result.Errors = append(result.Errors, errFormatter(err.(error)))
 			}
 			select {
 			case out <- result:
@@ -44,17 +50,18 @@ func Execute(p ExecuteParams) (result *Result) {
 			}
 		}()
 		exeContext, err := buildExecutionContext(buildExecutionCtxParams{
-			Schema:        p.Schema,
-			Root:          p.Root,
-			AST:           p.AST,
-			OperationName: p.OperationName,
-			Args:          p.Args,
-			Result:        result,
-			Context:       p.Context,
+			Schema:               p.Schema,
+			Root:                 p.Root,
+			AST:                  p.AST,
+			OperationName:        p.OperationName,
+			Args:                 p.Args,
+			Result:               result,
+			Context:              p.Context,
+			CustomErrorFormatter: errFormatter,
 		})
 
 		if err != nil {
-			result.Errors = append(result.Errors, gqlerrors.FormatError(err))
+			result.Errors = append(result.Errors, errFormatter(err))
 			return
 		}
 
@@ -67,7 +74,7 @@ func Execute(p ExecuteParams) (result *Result) {
 
 	select {
 	case <-ctx.Done():
-		result.Errors = append(result.Errors, gqlerrors.FormatError(ctx.Err()))
+		result.Errors = append(result.Errors, errFormatter(ctx.Err()))
 	case r := <-resultChannel:
 		result = r
 	}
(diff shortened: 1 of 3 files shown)


ccbrown:
I think there are a lot of ways you can already accomplish this without any library changes:

* Modify your error types to have the correct extensions.
* Modify your resolvers to return error types with the right extensions.
* Modify your schema to convert errors to types with the right extensions:

    ```go
    func resolveWrapper(resolve graphql.FieldResolveFn) graphql.FieldResolveFn {
    	return func(p graphql.ResolveParams) (interface{}, error) {
    		v, err := resolve(p)
    		// modify v and err accordingly
    		return v, err
    	}
    }
    
    func WrapResolvers(schema *graphql.Schema) {
    	for typeName, t := range schema.TypeMap() {
    		if strings.HasPrefix(typeName, "__") {
    			continue
    		}
    		obj, ok := t.(*graphql.Object)
    		if !ok {
    			continue
    		}
    		for _, f := range obj.Fields() {
    			if f.Resolve != nil {
    				f.Resolve = resolveWrapper(f.Resolve)
    			}
    		}
    	}
    }
    ```

racerxdl:
Well, for sure. But the easier way is to make a function that can adapt the error as you want. That's also the default way for other graphql libraries in other languages (for example javascript).

ccbrown:
> That's also the default way for other graphql libraries in other languages (for example javascript).

Does the JavaScript implementation actually have this feature? If so, reference implementation parity would certainly be a good reason to add this.

racerxdl:
Yes, it does. https://graphql.org/graphql-js/express-graphql/#graphqlhttp
The GraphQL HTTP Client has a `formatError` function that is called before sending the error back to the client.

ccbrown:
> Yes, it does. https://graphql.org/graphql-js/express-graphql/#graphqlhttp
The GraphQL HTTP Client has a formatError function that is called before sending the error back to the client.

That's not actually in the core JavaScript GraphQL implementation ([graphql/graphql-js](https://github.com/graphql/graphql-js)); it's part of a separate, higher level library ([graphql/express-graphql](https://github.com/graphql/express-graphql)).

In the same way that it's not part of the core implementation for JavaScript, my opinion is that it shouldn't be part of the core implementation here either. It would be more comparable to your example if the functionality went into one of the other higher level Go packages such as [graphql-go/handler](https://github.com/graphql-go/handler) instead.

racerxdl:
That's true. I would implement on handler only if it was possible, but it wasnt.

ccbrown:
Maybe a simpler change that would enable this would be if `gqlerrors.FormattedError` just stored and provided access to the original error? That might also help meet other error-related needs such as #312.

racerxdl:
@ccbrown that might not solve the issue, since you have no access to that error before is sent back to the user, and adding original error directly to it might leak private information and / or break the GraphQLError format.

ccbrown:
If you add an unexported `originalError error` field and an `OriginalError() error` method to `FormattedError`, it wouldn't effect the serialization. So it's easy to avoid that kind of leak or breakage.

If that were added, you could then just add something like this to [graphql-go/handler](https://github.com/graphql-go/handler):

```go
if customFormatter := h.CustomErrorFormatter; customFormatter != nil {
    formatted := make([]gqlerrors.FormattedError, len(result.Errors))
    for i, err := result.Errors {
        formatted[i] = customFormatter(err.OriginalError())
    }
    result.Errors = formatted
}
```

racerxdl:
Hmm, thats true. Let me try to do that.

racerxdl:
Lol, now I need to check those errors. These weren't there XD

zonr:
Your change to add an additional field for tracking back the underlying error can also be found in #279 ([link](https://github.com/graphql-go/graphql/pull/279/commits/94e801b5e666b617605b0868e7589c6a0446922d#diff-8f1cf99b975e822516aeb1025851300f)). Those CI failures is because many tests use `reflect.DeepEqual` to compare the errors returning from  GraphQL execution (which will have `originalError` or `cause` being set) against the one created by hand for assertion (which doesn't and can't set the field) ([example](https://github.com/graphql-go/graphql/blob/master/directives_test.go#L86-L92)). #279 addressed this issue by adding custom compare functions ([link](https://github.com/graphql-go/graphql/pull/279/commits/a31b4b3ccbb6064e1832ee691e5ffb4c9bccbb45#diff-d85486dab010c333b6e0880a348f537c)) for checking `graphql.Result` and `gqlerrors.FormattedError` in tests ([link](https://github.com/graphql-go/graphql/pull/279/commits/a31b4b3ccbb6064e1832ee691e5ffb4c9bccbb45)).

1046102779:
it may use [upgrade handler](github.com/1046102779/handler), [src](github.com/graphql-go/handler)

image

image

maybe try it!

demo: 

image

limoli:
Thank you @1046102779 !
Have you already send a pull request about this?
It would be useful to have in the official version.

limoli:
Any news?

PR #387, opened 2018-08-26
Add a new Duration scalar

pseudomuto:
@chris-ramon 

Similar to what was done for `DateTime` objects, I've added a new scalar type (`Duration`) to handle ser(de) of `time.Duration` values. The values are serialized to the string form of Duration (e.g. `2h34m5s900ns` or `1h30m0s`). 

Durations can be deserialized from `string`, `int64`, and `float64` types which makes these all passable as input values (all equivalent):

* `"1h30m15s"`
* `int64(5415000000000)`
* `float64(5415.000000)`

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -566,3 +566,51 @@ var DateTime = NewScalar(ScalarConfig{
 		return nil
 	},
 })
+
+var Duration = NewScalar(ScalarConfig{
+	Name:        "Duration",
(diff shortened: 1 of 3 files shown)


pseudomuto:
> Types like those would probably be better off in a sub-package or separate repo.

I agree with that. I noticed the DateTime one and thought maybe someone else needs Duration as well. But seeing as neither are part of the graphql spec, I think it makes sense to close this PR.

Thanks for the feedback!

Issue #178, opened 2017-01-05
Support null literal


jonricaurte:
The GraphQL language added null literal support a few months ago and it has not been added to graphql-go:

https://github.com/facebook/graphql/pull/83

seriallink:
+1

bbuck:
@seriallink That was an accident on my part, I'm sorry. The body of the comment was:

> +1

But in the future, please refrain from commenting things like that (as there is the reactions you can add to other comments (as of this comment here 8, including myself, have reacted with +1 and/or thumbs up).

MTigra:
It is still not fixed:(
https://stackoverflow.com/questions/49422706/golang-graphql-fails-to-tell-nil-pointer-from-nil-literal

david4shure:
Any movement on this? @chris-ramon @sogko @lenaten

mpenick:
Added a fairly well tested PR for this https://github.com/graphql-go/graphql/pull/536.

PR #401, opened 2018-09-23
Add support for null literals

evulse:
Choice of implementation was based on not breaking existing code. For instance the following will not break with this change
```go
price, priceOk := params.Args["price"].(float64)
```
However you can now detect null by adding another type assertion
```go
price, priceOk := params.Args["price"].(float64)
_, priceNull := params.Args["price"].(graphql.NullValue)
```
As you can see the value of `graphql.NullValue` is irrelevant, we just need to test for its existence.

The same thing can be accomplished by using a type switch, which there is an example of in the new example `crud-null`

The motivation for this change is to bring inline with spec and that forcing null values on database fields is rather common. Plus obviously needed this functionality for current application use.

Also addresses https://github.com/graphql-go/graphql/issues/178


ccbrown:
@evulse Glad to see someone working on support for this! One question though: Why adding is a new type to represent null values necessary? Couldn't we just do this using `nil`?

```go
func main() {
	m := map[string]interface{}{}

	_, priceOk := m["foo"].(float64)
	v, priceNull := m["foo"]
	fmt.Println(priceOk, priceNull && v == nil)

	m["foo"] = nil

	_, priceOk = m["foo"].(float64)
	v, priceNull = m["foo"]
	fmt.Println(priceOk, priceNull && v == nil)
}
```

```
false false
false true
```

– https://play.golang.org/p/tXyzbrMsXHy

evulse:
@ccbrown I guess thats completely possible, I guess it was more to avoid any possible situations of the value being nil. Instead of going through and confirming that nil is never possibly outputted from maybe a nil pointer or struct it was easier to just add a new type knowing that nothing could possibly be using the new type.

clery:
Any news on this pull request ?

cettoana:
Facing a problem and I think this pull request fixes that,
any updates of this PR?

jorgebay:
@chris-ramon Do you have some time to review this PR?

What can I  do to help getting this patch merged?

seriallink:
Are there any chances this pull request will be merged into master? Is there anything we can do to help?

PR #403, opened 2018-10-02
exposing Path in ResolveInfo to follow the reference implementation

bhoriuchi:
This pull request exposes the response path in the ResolveInfo as Path to match the reference implementation (https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L792-L803) and changes ResponsePath to an exported struct.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -582,6 +582,7 @@ type FieldResolveFn func(p ResolveParams) (interface{}, error)
 type ResolveInfo struct {
 	FieldName      string
 	FieldASTs      []*ast.Field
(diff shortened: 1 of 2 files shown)


quangbuule:
This is actually what I need, I created issue yesterday without knowing there is a PR from you. Seems no-one gives a shit about this.

quangbuule:
https://github.com/graphql-go/graphql/issues/404

chris-ramon:
LGTM 👍 – Thanks a lot @bhoriuchi!

Issue #402, opened 2018-09-26
Panics are treated like exceptions, causing information leakage


liamg:
When a resolver panics, the current implementation:

a. Recovers from the panic
b. Returns the error from the panic to the user in the GraphQL error list

See https://github.com/graphql-go/graphql/blob/master/executor.go#L38

Panics in go are meant to be game over, cannot continue, so if a critical issue is encountered where the program cannot continue, the last thing I want to do is continue AND leak details of the panic to my end users.

Can we at least make panic recovery optional, or ideally remove it altogether?

Issue #243, opened 2017-09-11
Custom panic handler


drimpact:
When a resolver panics, graphql recovers and puts the error into the generated graphql, which is really nice.

I would like to capture these panics to an external system so that I can identify issues with my code. Is there anyway to provide a function which gets called with the captured panics? If not, do you think it is a common use case and hence worth adding?

Just to be clear I don't think it should change how graphql-go handles the panic, it just allows people to record the panic happened

rodrigo-brito:
It will help a lot. 👍

Niondir:
I have the same request, I like to add file and linke numbers to my errors returned from panics.

Issue #28, opened 2015-10-26
How do you throw errors?


pyros2097:
I have a new graphql scalar type
```go
types.NewGraphQLScalarType(types.GraphQLScalarTypeConfig{
		Name: name,
		Serialize: func(value interface{}) interface{} {
			return value
		},
		ParseValue: func(value interface{}) interface{} {
			return value
		},
		ParseLiteral: func(valueAst ast.Value) interface{} {
			log.Info("Parsing Literal" + valueAst.GetKind())
			return errors.new("tester")
                        //panic("tester") works
		},
```
I construct an error and send it but in the response I don't get any error. But if I `panic` it seems to work. But isn't this not idiomatic go. Shouldn't the functions return an `(interface{}, error)`

bbuck:
I've noticed a similar issue with errors when passing `nil` as the value for a `types.GraphQLNonNull(...)` which should throw an error.

sogko:
Hi guys,

Currently user functions (such as `ParseLiteral`, `ResolveFn` etc) that are passed into configs and executed are caught/recovered by the `executor`. It will then be returned as `Errors` in `GraphQLResult` struct.

```go
type GraphQLResult struct {
	Data   interface{}                           `json:"data"`
	Errors []graphqlerrors.GraphQLFormattedError `json:"errors,omitempty"`
}
```

As to why the current implementation uses `panic` instead of returning a more idiomatic `error` value, I chose to defer from making that decision, instead made a straight port from `graphql-js` which throws error everywhere and anywhere, and wait to hear from the community :)

Personally I would love to see a move to a more go-idiomatic library as the project shapes itself, so I would vote for `(interface{}, error)` return signature for user functions as well.

@bbuck That's interesting. Could you share the JSON response from the query? (If this is off-topic, maybe in a separate issue)

Thanks guys!

bbuck:
I definitely think a move to a more idiomatic approach would be better. As far as I understood the graphql-js implementation was that it was only created to server as an example of how you could implement it. I'm pretty sure making the necessary alterations to bring a GraphQL server library to Go would make sense to apply Go processes to it.

[...]

PR #413, opened 2018-10-16
Adopt idiomatic error handling for executor.

zonr:
Stop using `panic`/`recover` to throw errors. Instead, stick to Go's
style for handling error. Function reports errors by returning an Go's
`error` value (instead of passing with `panic`). Errors are explicitly
checked or bypassed at where they might occur (instead of checking the
return value from `recover`).

Discussion: #28
Related: #44
Invalidate: #243, #253
Close: #402


atombender:
What's preventing this from being merged?

Fontinalis:
@egonelbre Could you take a look at this?

executor.go
```diff
@@ -42,20 +42,19 @@ func Execute(p ExecuteParams) (result *Result) {
...
 			result.Errors = append(result.Errors, extErrs...)
 		}
-
 		addExtensionResults(&p, result)
 	}()
 
 	resultChannel := make(chan *Result, 2)
 
-	go func() {
+	go func(out chan<- *Result, done <-chan struct{}) {
```
egonelbre: nit: no need to pass them in explicitly it's easier to see when using `resultChannel` directly.

executor.go
```diff
@@ -42,20 +42,19 @@ func Execute(p ExecuteParams) (result *Result) {
...
 		result := &Result{}
 
 		defer func() {
```
egonelbre: by not using recover `defer func() { resultChannel <- result }()`  will suffice

executor.go
```diff
@@ -79,11 +77,12 @@ func Execute(p ExecuteParams) (result *Result) {
 			Root:             p.Root,
 			Operation:        exeContext.Operation,
 		})
-	}()
+		return
```
egonelbre: unneeded extra return

atombender:
Can this get some attention?

atombender:
@Fontinalis @egonelbre What's left for this to be merged? It's been 7 months, and this is bugging us.

egonelbre:
@Fontinalis I made the readability changes I commented in https://github.com/egonelbre/graphql/commit/3e383bec6e59522df4bed6219cf4f2ae11c0fe73, but I have no clue how to get these changes into this PR.

We could always merge as is and I can a do separate PR against.

jpascal:
How long time we must wait to this feature?

PR #423, opened 2018-11-28
errors: adds OriginalError support

chris-ramon:
#### Overview
- Build on top of: https://github.com/graphql-go/graphql/pull/379 — Thanks a lot @racerxdl! :+1: 

- This PR adds support for retrieving the original error via `FormattedError.OriginalError()`.

  - No breaking changes.

  - Does not expose original errors when JSON serializing since it is available only by a method and the internal field `originalError` is unexported.

- Matches `graphql-js` reference implementation `originalError` (naming and private variable):
  ```javascript
  declare class GraphQLError extends Error {
    constructor(
      message: string,
      nodes?: $ReadOnlyArray<ASTNode> | ASTNode | void | null,
      source?: ?Source,
      positions?: ?$ReadOnlyArray<number>,
      path?: ?$ReadOnlyArray<string | number>,
      originalError?: ?Error, <---------------------------------------------------------
      extensions?: ?{ [key: string]: mixed },
    ): void;
  ```
  Ref: [Link](https://github.com/graphql/graphql-js/blob/f0ae3f4c0ba469c8f2147f8d12c9e670313b865c/src/error/GraphQLError.js#L29)

#### Test plan
- Unit tests.
- Adds tests for handling errors from resolver.
  - Returning an `errors.New(...)` error.
  - Returning a custom error.
  - Returning a custom error with extensions.
  - Panic from resolver.

### DIFF
--- a/abstract_test.go
+++ b/abstract_test.go
@@ -491,6 +491,7 @@ func TestResolveTypeOnInterfaceYieldsUsefulError(t *testing.T) {
       }
     }`
 
+	originalError := gqlerrors.NewFormattedError(`Runtime Object type "Human" is not a possible type for "Pet".`)
 	expected := &graphql.Result{
 		Data: map[string]interface{}{
 			"pets": []interface{}{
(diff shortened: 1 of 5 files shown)


limoli:
When are you going to merge this? :) @chris-ramon

PR #428, opened 2018-12-10
trace support

chris-ramon:


Fontinalis:
@chris-ramon When will this land in the master branch? Also, any help needed with this one? Happy to help

linniksa:
Can you look at https://github.com/apollographql/apollo-tracing extension , it more yousefull in dev environment
maybe allow trace parse/validate stage?
is current Tracer interface allow provide "path" field as described in apollo-tracing extension?

Fontinalis:
@chris-ramon I created a PR #448 that would add support for extensions. 
@linniksa In my PR, there's an example (`star-wars-with-trace') that has a simple tracer, that basically adds the support for the "apollo-tracing". The field's resolve info can be viewed from an extension, so the full "path" field could be viewed.

PR #431, opened 2019-01-03
Fix panic when mutations or subscriptions is not configured

linniksa:
when call 
```graphql
mutation { a }
``` 

on schema that mutations is not configured, i got:

```json
{
  "data": null,
  "errors": [
    {
      "message": "runtime error: invalid memory address or nil pointer dereference",
      "locations": []
    }
  ]
}
```
this PR fix that


chris-ramon:
LGTM :+1: It aligns with the [graphql-js reference implementation](https://github.com/graphql/graphql-js/blob/fbd9c4700a3dd6d72c9190d25d0e5c223dbfe444/src/utilities/getOperationRootType.js#L35-L52). — Thanks a lot @linniksa!

PR #432, opened 2019-01-03
Add extensions to Result

linniksa:
add extensions key to result as mentioned in specification https://facebook.github.io/graphql/June2018/#sec-Response-Format

also may be make Errors a first member of Result as proposed in spec?

### DIFF
--- a/types.go
+++ b/types.go
@@ -4,11 +4,10 @@ import (
 	"github.com/graphql-go/graphql/gqlerrors"
 )
 
-// type Schema interface{}
-
 type Result struct {
-	Data   interface{}                `json:"data"`
-	Errors []gqlerrors.FormattedError `json:"errors,omitempty"`
+	Data       interface{}                `json:"data"`
+	Errors     []gqlerrors.FormattedError `json:"errors,omitempty"`
+	Extensions map[string]interface{}     `json:"extensions,omitempty"`
 }
 
 func (r *Result) HasErrors() bool {


chris-ramon:
Hi @linniksa, thanks a lot for your contribution.

- Yes, it is correct GraphQL spec. allows us to use `extensions` in the GraphQL response.
- Following the `graphql-js` reference implementation, it does not include the `extensions` field itself in the [`ExecutionResult`](https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L110-L119):

  ```javascript
  /**
   * The result of GraphQL execution.
   *
   *   - `errors` is included when any errors occurred as a non-empty array.
   *   - `data` is the result of a successful execution of the query.
   */
  export type ExecutionResult = {
    errors?: $ReadOnlyArray<GraphQLError>,
    data?: ObjMap<mixed>,
  };
  ```
- Meaning `extensions` can be included but is up to the end-user to do so, for example the node lib [`express-graphql`](https://github.com/graphql/express-graphql) can receive an [optional function `extensions`](https://github.com/graphql/express-graphql#options) which is invoked to include its response in the graphql's response:

  > An optional function for adding additional metadata to the GraphQL response as a key-value object. The result will be added to "extensions" field in the resulting JSON. This is often a useful place to add development time metadata such as the runtime of a query or the amount of resources consumed. This may be an async function. The function is given one object as an argument: { document, variables, operationName, result, context }.

- A very similar lib to express-graphql is [`github.com/graphql-go`](https://github.com/graphql-go/handler) currently does not support extensions but you if could file an issue for it and we might implement it in the near feature.

- In regards including errors in the top response, I think is a very similar case than the just described above is up to the end-user. The [github.com/graphql-go/handler](https://github.com/graphql-go/handler) lib has support for this use case via the `FormatErrorFn` here is [an example for it](https://github.com/graphql-go/handler/blob/d08c80956e91ac6d3b0679ac85af7397bef842c3/handler_test.go#L270-L280).

chris-ramon:
Closing this one since effectively `extensions` can be added as part of the response, but is up to the library user, and it [is not part of the execution result](https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L110-L119).

PR #433, opened 2019-01-03
simplify lexer.TokenType

linniksa:
This also speeds up lexer a bit


chris-ramon:
Hi @linniksa, thanks a lot for your contribution!

LGTM :+1: — This is a neat improvement! def. better to use constant integers rather than accessing map values performance wise. :rocket: :100:

Issue #429, opened 2018-12-17
Potential data race


BurtonQin:
I found a potential data race between the following two instructions:
1. https://github.com/graphql-go/graphql/blob/21fe71deb5864761ea88e1dbc98e973964c9127c/executor.go#L57
2. https://github.com/graphql-go/graphql/blob/21fe71deb5864761ea88e1dbc98e973964c9127c/executor.go#L70

songlh:
Any update? By reading the code, this is definitely a data race.

PR #434, opened 2019-01-08
Add Result.AppendErrors

fishy:
This is a thread-safe way of appending errors into Result.Errors, as we
are currently doing it in different goroutines.

This fixes https://github.com/graphql-go/graphql/issues/429


songlh:
Should we merge this pull request?

chris-ramon:
LGTM :+1: :ship: — Thanks a lot for your contribution @fishy, and for your review @songlh!

PR #441, opened 2019-02-06
Fix panic when a blockstring has only 1 line and blank

miiton:
When I send a blockstring that has only 1 line and blank...

```
""""""
```

`language/lexer.go` panics.

```
panic: runtime error: index out of range [recovered]
        panic: runtime error: index out of range
```

This PR fix that.


chris-ramon:
Thanks a lot for fixing it @miiton :+1: , closing this one in favor of: https://github.com/graphql-go/graphql/pull/456

Issue #457, opened 2019-03-10
Cannot send null as argument for scalar type


clery:
Hi,

I'm using graphql v0.7.7 and I noticed I cannot send null in a scalar argument.

It's really easy to reproduce, just create a field that takes a string as an argument, and try sending null, it should return something like `Argument "some_name" has invalid value <nil>. Expected type "String", found <nil>."`

clery:
I believe pull request #401 fixes this.

PR #460, opened 2019-03-10
Add NullValue to allow sending null in a value literal

clery:
I noticed `null` was not a valid value to return anywhere in a ParseLiteral, so I took a look at the way the parsing was done with graphql-js and tried reimplementing their NullValue.

There may be other stuff to do, for example fixing #457 to allow sending null to any scalar type as an argument.

Feel free to tell me if anything looks bad or whatever 🙂


clery:
Just noticed that there already is a pull request about this #401

PR #463, opened 2019-03-16
Compare all args when testing for overlap

bencallaway:
Testing for overlapping arguments was only correctly testing fields with a single argument, and invariably failing for fields with multiple arguments. I've added an additional test for comparing two fields w/ two identical arguments, and a tiny fix to the `sameArguments()` function within `rules_overlapping_fields_can_be_merged.go`.


chris-ramon:
@bencallaway thanks a lot! 👍

Closing this one since we merge this improvement via: https://github.com/graphql-go/graphql/pull/521

PR #465, opened 2019-03-23
Implement Object.Definition()

KevinShiCA:
The `Object.Definition()` method was stubbed to return an empty string. This PR makes it return `gt.PrivateDescription`.


siliconbrain:
I just ran into this too, wrote the same patch, then found this PR before opening my own. Please, merge this! It's a one liner, the author even wrote a test for it.

PR #475, opened 2019-04-10
Adds support for custom marshalers through the encoding.TextMarshaler interface

### LINKED ISSUE #471: Bad JSON marshaling for custom Marshaler
I am using the library `github.com/guregu/null` in order to provide an easy management of null fields on my database entities. The library provides the following structures:
- `null.String`
- `null.Int`
- `null.Bool`
- `null.Time`

Every structure implements **JSONMarshaler** and **JSONUnmarshaler** interfaces, so a simple `json.Marshal`returns the expected JSON output.

_Expected Result (Simple_ `json.Marshal`_)_
```json
{
    "mobile": "1234567891"
}
```

_Current Result (GraphQL response)_
```json
{
    "mobile": "{{1234567891 true}}"
}
```

I imagine that this happens because the structure has two exported fields (Value and Valid).
But this **should not happen** since the structure implements the JSON marshaler and unmarshaler interfaces.

I think that the problem is in `scalars.go` file. Take String scalar as example. The `coerceString` casts the string to a string pointer when the interface implements the Marshaler interface.
```go
func coerceString(value interface{}) interface{} {
	if v, ok := value.(*string); ok {
		if v == nil {
			return nil
		}
		return *v
	}
	return fmt.Sprintf("%v", value)
}

// String is the GraphQL string type definition
var String = NewScalar(ScalarConfig{
	Name: "String",
	Description: "The `String` scalar type represents textual data, represented as UTF-8 " +
		"character sequences. The String type is most often used by GraphQL to " +
		"represent free-form human-readable text.",
	Serialize:  coerceString,
	ParseValue: coerceString,
	ParseLiteral: func(valueAST ast.Value) interface{} {
		switch valueAST := valueAST.(type) {
		case *ast.StringValue:
			return valueAST.Value
		}
		return nil
	},
})
```
I don't know what is the best practice for this library, but I would like to get help about this and provide a smart solution.

Fontinalis:
This PR adds a check to each built-in coerce function, so if the value implements the `encoding.TextMarshaler` interface, it'll be marshaled and converted from that, so custom types, like some nullable type from `github.com/guregu/null` would work correctly.

Closes #471

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -1,6 +1,7 @@
 package graphql
 
 import (
+	"encoding"
 	"fmt"
 	"math"
 	"strconv"
@@ -15,6 +16,12 @@ import (
 // n.b. JavaScript's integers are safe between -(2^53 - 1) and 2^53 - 1 because
 // they are internally represented as IEEE 754 doubles.
 func coerceInt(value interface{}) interface{} {
+	if v, ok := value.(encoding.TextMarshaler); ok {
+		bs, err := v.MarshalText()
+		if err == nil {
+			value = string(bs)
+		}
+	}
 	switch value := value.(type) {
 	case bool:
 		if value == true {
@@ -168,6 +175,12 @@ var Int = NewScalar(ScalarConfig{
 })
 
 func coerceFloat(value interface{}) interface{} {
+	if v, ok := value.(encoding.TextMarshaler); ok {
+		bs, err := v.MarshalText()
+		if err == nil {
+			value = string(bs)
+		}
+	}
 	switch value := value.(type) {
 	case bool:
 		if value == true {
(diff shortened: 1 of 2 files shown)


billkillapp:
Could we merge this feature ASAP?

Fontinalis:
Waiting for at least one review.. I could merge it if it would be someone else’s PR, but cannot do it without a review since it’s mine..

limoli:
When are you going to merge this important feature into master?

billkillapp:
Hi Andrea,

I got this mail. What does it mean? :-)


*Sonni Christine Jakobsen*
*Co-founder & CEO*

+47 466 60 626
[image: image.png]
https://billkill.no/


fre. 11. okt. 2019 kl. 12:20 skrev Andrea Limoli <notifications@github.com>:

Fontinalis:
@chris-ramon Can you review this?

PR #476, opened 2019-04-11
Fix incorrect context hierarchy when using extensions

sfriedel:
When using an extension that creates new child contexts in `ResolveFieldDidStart` the resulting context hierarchy is wrong.

Given the query

```
query {
  a {
    foo
    bar
    baz
  }
}
```

the following context hierarchy would be created

```
a
|
+-> foo
    |
    +-> bar
        |
        +-> baz
```

So each context has the context of the previously executed field as a parent context.

This PR changes this so that the context in each field has the parent context of the parent field:

```
a
|
+-> foo
|    
+-> bar
|
+-> baz
```

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -111,6 +111,14 @@ type executionContext struct {
 	Context        context.Context
 }
 
+// WithContext creates a shallow copy of the executionContext and replaces the embedded context.Context of the copy with ctx
+func (eCtx *executionContext) WithContext(ctx context.Context) *executionContext {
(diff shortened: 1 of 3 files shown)


sfriedel:
closing this PR for the now because we noticed that it breaks error propagation in some cases

Fontinalis:
Sorry to hear, I was just about to look into the PR. Thanks Sebastian for starting to work on it! 

I will look into it anyway, for some inspiration at least

PR #489, opened 2019-06-14
Fix coercion of typedef primitives and their pointers

### LINKED ISSUE #488: Aliased pointer types become pointer addresses
Test case:

```go
package main

import (
	"context"
	"encoding/json"
	"fmt"
	"log"

)

type Value string

func main() {
	s, err := graphql.NewSchema(graphql.SchemaConfig{
		Query: graphql.NewObject(graphql.ObjectConfig{
			Name: "RootQuery",
			Fields: graphql.Fields{
				"value": &graphql.Field{
					Type: graphql.String,
						v := Value("hello")
						return &v, nil
					},
				},
			},
		}),
	})
	if err != nil {
		log.Fatal(err)
	}

		Schema:        s,
		RequestString: "query MyQuery { value } ",
		Context:       context.TODO(),
	})

	fmt.Printf("%#v\n", result.Data)
}
```

This prints

```go
map[string]interface {}{"value":"0xc00000f500"}
```

instead of the expected:

```go
map[string]interface {}{"value":"hello"}
```

Annoyingly, there's no way to work around this by implementing `MarshalJSON` on `Value`, since it seems graphql-go doesn't use it.

I'll see if I can create a PR.

Related to #199.

atombender:
For example, with `type MyInt int`, values of type `MyInt` and `*MyInt` should be treated as ints. Without this fix, they return the pointer value, e.g. `0xc00000f500`. Fixes #488.

### DIFF
--- a/scalars.go
+++ b/scalars.go
@@ -3,11 +3,78 @@ package graphql
 import (
 	"fmt"
 	"math"
+	"reflect"
 	"strconv"
 	"time"
 
 	"github.com/graphql-go/graphql/language/ast"
 )
+
+func unwrapInt(value interface{}) (interface{}, bool) {
+	r := reflect.Indirect(reflect.ValueOf(value))
+	if !r.IsValid() || (r.Kind() == reflect.Ptr && r.IsNil()) {
+		return nil, false
+	}
+
+	switch r.Kind() {
+	case reflect.Int:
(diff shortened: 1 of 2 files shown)


atombender:
@sogko @chris-ramon @Fontinalis Review, please?

Issue #145, opened 2016-07-07
wrong execution order of multiple mutations in one request


sfriedel:
The GraphQL spec states that mutations should be executed serially and in the order that they were specified.

Currently this isn't the case in graphql-go. Mutations are executed in random order (I'd guess they are stored in a map somewhere because order changes with every restart).

> When evaluating a grouped field set serially, the executor must consider each entry from the grouped field set in the order provided in the grouped field set. It must determine the corresponding entry in the result map for each item to completion before it continues on to the next item in the grouped field set:

https://facebook.github.io/graphql/#sec-Serial-execution

robermorales:
Something new about this?

djmcgreal:
The problem is in `executeFields(Serially)`, `p.Fields` is a map created by `collectFields`. Since Go iterates over maps in an arbitrary order, this isn't an approach that can be compatible with the spec.

PR #493, opened 2019-07-03
adds a field sort by source location before executing serially

bhoriuchi:
Resolves https://github.com/graphql-go/graphql/issues/145 by sorting fields by source location before executing serially

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"sort"
 	"strings"
 
 	"github.com/graphql-go/graphql/gqlerrors"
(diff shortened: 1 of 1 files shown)


djmcgreal-cc:
I'd love this to get some attention from reviewers. Not having mutations executed in the correct order means that we go back to needing multiple operation requests to enforce order of a chain of mutations.

Personally, I'd also like to see this configurable to be usable outside of the top level of a mutation operation. Our schema wraps every operation in an `authenticate` field so it's actually the second level of fields we would like executed in serial.

Perhaps there should also be a test case for this.

chris-ramon:
@bhoriuchi thanks a lot for improving the serially execution order! –– it definitely looks good to me.

I've tested it and seems working neat! –– Sharing the gist with the working example: https://gist.github.com/chris-ramon/6c98e31259d1c7837fc3f240ee4f988d

chris-ramon:
> I'd love this to get some attention from reviewers. Not having mutations executed in the correct order means that we go back to needing multiple operation requests to enforce order of a chain of mutations.
> 
> Personally, I'd also like to see this configurable to be usable outside of the top level of a mutation operation. Our schema wraps every operation in an `authenticate` field so it's actually the second level of fields we would like executed in serial.
> 
> Perhaps there should also be a test case for this.

@djmcgreal-cc thanks a lot for the feedback! 👍 –– perhaps could you file an issue and share an example of the query your mentioning, so we could make this PR configurable via a follow-up PR.

Issue #242, opened 2017-09-11
Adding "Subscribe" field to subscription definitions


robbawebba:
I'm currently implementing a GraphQL subscription server and I'm interested in adding subscription logic to each subscription definition to use with my PubSub manager. **Is it possible to add a custom field to a subscription definition that I can use to pass data/functions around?** If not, would the core team consider this as a possible enhancement to the way subscriptions work?

The idea is to have a field similar to `Resolve` (possiblly called `Subscribe`) that will be accesible by any PubSub system that contains a function. This function would be responsible for the logic that defines how a subscription listens for events based on the client's query and publishes updates to the client. It's implementation and requirements, similar to the `Resolve` function, would be pretty relaxed and would rely on the developer to fill the function with the logic that they need to work with their data layer. So when a developer is definining subscriptions and needs a place to create query-specific logic for listening for events and publishing changes, they will be able to define that along side the rest of the subscription definition. 

Lastly, the PubSub manager will be able to use this function when creating a new instance of a subscription. 

Below is an example of what I'm trying to accomplish. This particular implementation of the proposed `Subscribe` field listens for changes to a RethinkDB database based on the `contactID` provided from the client's subscription query.  After the change listener is set up, When a change to the database is received, the changes are published to the PubSub manager.
```
type dbChange struct {
	NewVal interface{} `gorethink:"new_val"`
	OldVal interface{} `gorethink:"old_val"`
}

var rootSubscription = graphql.NewObject(graphql.ObjectConfig{
	Name: "rootSubscription",
	Fields: graphql.Fields{
		"contactChanged": &graphql.Field{
			Type: contact,
			Args: graphql.FieldConfigArgument{
				"contactID": &graphql.ArgumentConfig{
					Type: graphql.String,
				},
			},
			Resolve: func(p graphql.ResolveParams) (interface{}, error) {
				return p.Info.RootValue, nil
			},
			Subscribe: func(params graphql.ResolveParams) error {
				ID := params.Args["contactID"].(string)
				contactCursor, err := gorethink.DB("ContactsApp").
					Table("Contacts").
					Get(ID).
					Changes().
					Run(db.Session)
				if err != nil {
					fmt.Errorf(`Error reading database: %v`, err)
					return err
				}
				contactChanges := make(chan dbChange)
				contactCursor.Listen(contactChanges)
				fmt.Println("Listening for changes")

				for change := range contactChanges {
					subManager.publish("CONTACT_CHANGED", change)
				}
			},
		},
[...]

PR #495, opened 2019-07-15
Subscription execution

bhoriuchi:
Adds subscription execution support and a Subscribe resolver to Field and FieldDefinition types

Resolves https://github.com/graphql-go/graphql/issues/242


remorses:
Is the done channel necessary? Can’t you just close the channel or cancel the context?

bhoriuchi:
Probably not. I should probably revisit the strategy on the iterator here.

remorses:
I looked at the other graphql go libraries implementation and i found aout a few things:

-  If the `Subscribe` function is a method of `schema` then schema will implement [this interface](https://github.com/graph-gophers/graphql-transport-ws/blob/40c0484322990a129cac2f2d2763c3315230280c/graphqlws/internal/connection/connection.go#L53) so we can reuse the [graphql-transport-ws](https://github.com/graph-gophers/graphql-transport-ws) handler implementation. This handler already implements the websockets messaging logic

- There is no need for a different `Subscribe` function for every type, types under the `Subscription` object will return a channel and error in the `Resolver` function, this way we can have a similar UX as `gqlgen` and `graphql-gophers`, here is an [example subscription resolver](https://github.com/graph-gophers/graphql-go/blob/dae41bde9ef91c12e78863f0299348612f2d6214/subscription_test.go#L38)
- no need for a `done` channel, we can use the context `cancel` method to end the subscription, [here](https://github.com/graph-gophers/graphql-go/blob/dae41bde9ef91c12e78863f0299348612f2d6214/subscription_test.go#L38) is an example

bhoriuchi:
Thanks for the research on this. One of the goals of this project is to follow the official reference implementation. That is the motivation behind some of the existing design choices. I will review these. 

Using the context to cancel seems straightforward enough and I was originally using the context to cancel but ran into issues that I cant recall so I will revisit that.

bhoriuchi:
> * There is no need for a different `Subscribe` function for every type, types under the `Subscription` object will return a channel and error in the `Resolver` function, this way we can have a similar UX as `gqlgen` and `graphql-gophers`, here is an [example subscription resolver](https://github.com/graph-gophers/graphql-go/blob/dae41bde9ef91c12e78863f0299348612f2d6214/subscription_test.go#L38)

Please see graphql spec https://spec.graphql.org/June2018/#sec-Source-Stream and https://spec.graphql.org/June2018/#sec-Response-Stream. In a subscription the purpose of subscribe is to generate the source stream who's messages are used as the source in the resolve function to generate the response. 

> * no need for a `done` channel, we can use the `cancellable context` to end the subscription, the resolver will check if the context is cancelled to terminate the goroutine, [here](https://github.com/graph-gophers/graphql-go/blob/dae41bde9ef91c12e78863f0299348612f2d6214/subscription_test.go#L38) is an example

probably fine, I'd like to re-work the iterator anyway.

bhoriuchi:
@remorses I have changed the following
* Subscribe now takes a context as its first argument. Cancelling this context ends the subscription.
* Subscribe returns a `chan *Result`. This remove the use of the ResultIterator type which has been removed from the source code
* Subscribe function on a type is now expected to return a `chan interface{}`, this removes the use of the Subscriber type which has been remove from the source code

remorses:
I made some changes to your fork, i opened a pull request on your branch

remorses:
One thing that still does not work:

When omitting the `Resolve` function, the returned data is null, i think it is because the subscribe function should return the same shape of the query to make the default resolver work

bhoriuchi:
@chris-ramon would be great to get a review on this.

bhoriuchi:
Happy belated 1 year anniversary PR!

mtryfoss:
Any progress here? Seems like a nice feature.

exapsy:
This feature looks great indeed. It solves one of the problems with subscription. Do you guys feel comfortable merging this PR?

exs1st:
Hey. There is information when it will be merged?

bhoriuchi:
> This feature looks great indeed. It solves one of the problems with subscription. Do you guys feel comfortable merging this PR?

I believe so, just waiting for one of the project owners to review and PR. I don't believe there are any breaking changes in any of the code here. Adding a subscribe field is the only modification to object types and the subscription code is all self contained/new code.

pckilgore:
I agree that I would love to see this on master.  Has anyone reached out directly to @chris-ramon (seems to be the only active maintainer) to see if he is aware of this PR.  At this point, the most helpful thing would be reasons from him why it isn't being merged / reviewed.

bhoriuchi:
@pckilgore I have reached out and he has seen the PR.

jpascal:
Any progress here?

bhoriuchi:
> Any progress here?
> 
> I have one question to @bhoriuchi
> 
> If I have defined schema like:
> 
> ```
> schema {
>     ...
>     subscription: Subscriptions
> }
> type Subscriptions {
>    worked(arg: String): String
>    foo: Foo
> }
> type Foo {
>    foo(arg: Strring): String
>    bar: Bar
> }
> type Bar {
>    bar(arg: Strring): String
> }
> ```
> 
> Will it work?
> 
> ```
> subscription { worked(arg: "test") }
> ```
> 
> ```
> subscription { foo { foo(arg: "test") } }
> ```
> 
> ```
> subscription { foo { bar { bar(arg: "test") } } }
> ```
> 
> I have troubles with last two examples.

please elaborate on what the trouble is and provide full code including handlers.

bhoriuchi:
I found an issue with field selections returning nil. Working on a fix. Please do not merge yet. TY

bhoriuchi:
Never mind. This is working as designed. see https://github.com/apollographql/graphql-subscriptions/issues/51 . The even message needs to be nested in the subscription field name for it to resolve correctly. Merging can proceed

bhoriuchi:
Additionally here is an example of a create watcher function that can handle the nesting conditionally. it seems that when the subscription field returns say a scalar it does not require the additional step of nesting the value. This assumes theoretical event emitter

```go
func CreateWatcher(p graphql.ResolveParams, topic string) (interface{}, error) {
	c := make(chan interface{})

	off := DefaultEmitter.On(topic, func(msg ...interface{}) {
		data := msg[0]
		switch p.Info.ReturnType.(type) {
		case *graphql.Object, *graphql.Interface, *graphql.Union:
			data = map[string]interface{}{
				p.Info.FieldName: msg[0],
			}
		}

		c <- data
	})

	go func() {
		for {
			select {
			case <-p.Context.Done():
				off()
				close(c)
				return
			}
		}
	}()

	return c, nil
}
```

exapsy:
Any updates? I constantly see re-sync with master, but I dont see it ever actually merging with master. This has been going on like from 2019 :P What is going on? :D

AttilaTheFun:
@chris-ramon Is there anything still blocking this PR from merging? I see it has been kept alive but it's unclear what the status is at this time.

bhoriuchi:
For those of you following this thread, if you want to use this feature and don't mind using the replace feature in go mod, this is how i have been using it

```
replace github.com/graphql-go/graphql v0.7.9 => github.com/bhoriuchi/graphql v1.0.0-beta2
```

didrikmunther:
@chris-ramon Will this get merged? What is the status here?

bhoriuchi:
Happy belated 2 year anniversary PR!

asif-mahmud:
still waiting for the glorious day when this will be merged :)

chris-ramon (approved):
LGTM :+1: :ship:  — Thanks a lot for the great work & feedback everybody! All kudos to: @bhoriuchi & @remorses.

Confirmed this `graphql.Subscription` implementation follows the implementation from [`graphql-js`](https://github.com/graphql/graphql-js) and works as expected.

Also just for reference, I've put together a small working example integrating this `graphql.Subscription` implementation and GraphQL Playground: https://github.com/graphql-go/subscription-example

bhoriuchi:
Thanks for getting this in @chris-ramon can we get possibly a version (v0.7.10) tag applied?

chris-ramon:
> Thanks for getting this in @chris-ramon can we possibly get a version (v0.7.10) tag applied?

@bhriuchi great call! — Just created [v0.8.0](https://github.com/graphql-go/graphql/releases/tag/v0.8.0), released as minor version since we're adding a new end-user API: `graphql.Subscribe`.

Juvenal-Yescas:
Maybe someone needs this ping [example](https://gist.github.com/Juvenal-Yescas/9e976ea519a50227a65182cfae6d28ff)

PR #501, opened 2019-08-13
added sql-nullstring example

djangulo:
`database/sql` NullString implementation with a custom scalar type.


djangulo:
I'm glad you like it! Thank you for this amazing library!

Issue #510, opened 2019-09-17
Float misparsed as 32 bits instead of 64


lspgn:
Hello,
First I would like to thank you for this excellent library, I love using it.
I wanted to report a bug: a misparsed value:

I am passing 4294967290.
It is considered as an `*ast.IntValue` and therefore is parsed with a bit size of 32 (https://golang.org/pkg/strconv/#ParseFloat)
Example:
```
a, _ := strconv.ParseFloat("4294967290", 32)
fmt.Println(a)
// will return 4.294967296e+09
a, _ = strconv.ParseFloat("4294967290", 64)
fmt.Println(a)
// will return 4.29496729e+09 
```
My main problem is when I try to cast it: `uint32(4.294967296e+09)` will return 0.

I believe this could be fixed by setting a bit size of 64? But I'm not sure this is a bug.

You can try it:
https://play.golang.org/p/uVuUCtqrXV-

The line of code is:
https://github.com/graphql-go/graphql/blob/993e01c146460ccb5530f30a7d9ce19ee6e6c830/scalars.go#L299

Stack trace is:
```
github.com/graphql-go/graphql.glob..func4(0x1730f20, 0xc004515ad0, 0x8, 0x157f160)
    /hidden-path/vendor/github.com/graphql-go/graphql/scalars.go:296 +0xbd
github.com/graphql-go/graphql.(*Scalar).ParseLiteral(...)
    /hidden-path/vendor/github.com/graphql-go/graphql/definition.go:287
github.com/graphql-go/graphql.isValidLiteralValue(0x1734460, 0xc0001229a0, 0x1730f20, 0xc004515ad0, 0xa, 0x0, 0x1014553, 0x165c20f)
    /hidden-path/vendor/github.com/graphql-go/graphql/rules.go:1800 +0xba0
github.com/graphql-go/graphql.ArgumentsOfCorrectTypeRule.func1(0x15b62a0, 0xc004515b00, 0x157f160, 0xc0028fa3f0, 0x0, 0x0, 0xc00019b380, 0x7, 0x8, 0xc00019ba80, ...)
    /hidden-path/vendor/github.com/graphql-go/graphql/rules.go:77 +0xcf
github.com/graphql-go/graphql/language/visitor.VisitInParallel.func1(0x15b62a0, 0xc004515b00, 0x157f160, 0xc0028fa3f0, 0x0, 0x0, 0xc00019b380, 0x7, 0x8, 0xc00019ba80, ...)
[...]

Issue #165, opened 2016-10-16
Update precision for float64 parsing in graphql.Float


matheusmartin:
The current parsing of float values uses 32 precision which cause the resulting float variable from `graphql.ResolveParams` to have a dirty floating point, e.g., 32.45 would become 32.45000230230230.

I changes line 134 from

`if floatValue, err := strconv.ParseFloat(valueAST.Value, 32); err == nil {`

to

`if floatValue, err := strconv.ParseFloat(valueAST.Value, 64); err == nil {`

and it solved the problem.

bbuck:
Did the library only parse as 32 bit? The GraphQL spec specifies that floats should be handled as double precision. If there are more instances where floats are treated as `float32` instead of `float64` they should be changed.

PR #511, opened 2019-09-17
Fix scalar parse float issue #510

lspgn:
As listed in https://github.com/graphql-go/graphql/issues/510, this would fix the parsing issue.


chris-ramon:
LGTM :+1: — Thanks for fixing it @lspgn.

Just for the records there was a previous PR that partially fixed this: https://github.com/graphql-go/graphql/pull/334

Also based on [previous comments](https://github.com/graphql-go/graphql/issues/165#issuecomment-254975872) on related issues it validates that this PR's change is correct.

Issue #516, opened 2019-10-15
Update Schema


jcantonio:
Hello,
First thanks for this great project.
I ran into the issue when updating a schema.
For an update, I first remove the type then add the new one
```
delete(schema.TypeMap(), resourceType.Name())
schema.AppendType(resourceType)
```
It works when the schema doe not contain a field of type object.
But when using an object, it seems that updates have a collision name, as it returns the error:
> Schema must contain unique named types but contains multiple types named "objectA".
See code sample:
https://play.golang.org/p/N8iBmXzG-U1

Kind regards,

jcantonio:
Hi,
I just provided a pull request to fix it: 
https://github.com/graphql-go/graphql/pull/518
Cheers,
JClaude

PR #518, opened 2019-10-17
Fix name collusion when field is of type object

jcantonio:
This fixes the error 
> Schema must contain unique named types but contains multiple types named
when fields are of type Object.
See #516 
Cheers


bhoriuchi:
Tests not passing, probably won't be merged.

rxdrag:
> Tests not passing, probably won't be merged till they do.

any way to solve this problem?

PR #526, opened 2020-01-29
Always call resolveFieldFinishFn even if an error occurs in resolve.

purple4reina:
Hello! We at New Relic are currently working to create an Extension for tracing GraphQL calls using this package.

One thing that I have noticed is that if your resolve field function returns an error, the `graphql.ResolveFieldFinishFunc` for that field is never called. This means that users will not see any metrics or data about this error or this work done despite it clearly having happened.

```go
"erred": &graphql.Field{
    Type: graphql.String,
        // because an error is returned here, the graphql.ResolveFieldFinishFunc
        // is never called and no segments or metrics are created for this work
        return "", errors.New("ooops")
    },
},
```

The fix is pretty simple, just call the finish func before panicking the error. This is similar to what is done for the other Extension methods (parse, validate, and execute) which all call their finish functions even if an error occurs.

Please let me know what you think. I'm happy to make any changes you request.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -650,13 +650,13 @@ func resolveField(eCtx *executionContext, parentType *Object, source interface{}
 		Context: eCtx.Context,
 	})
 
-	if resolveFnError != nil {
-		panic(resolveFnError)
-	}
-
 	extErrs = resolveFieldFinishFn(result, resolveFnError)
 	if len(extErrs) != 0 {
 		eCtx.Errors = append(eCtx.Errors, extErrs...)
+	}
+
+	if resolveFnError != nil {
+		panic(resolveFnError)
 	}
 
 	completed := completeValueCatchingError(eCtx, returnType, fieldASTs, info, path, result)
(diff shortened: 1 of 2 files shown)


chris-ramon:
Thanks a lot @purple4reina –– neat to know that you are using the lib and working on a tracing extension.

It looks good to me since `ResolveFieldFinishFunc ` is mean to return the resolve function error if is the case:

https://github.com/graphql-go/graphql/blob/bed865f0458679ed5309c2cc4bfa82a2f84d8b5f/extensions.go#L26-L27

@Fontinalis could you give a look to this PR when you get a chance please ?

purple4reina:
Thanks a ton @chris-ramon! This change makes our lives so much easier and we really appreciate it! Any chance you know about when you'll be putting out another release?

chris-ramon:
good call @purple4reina, I've released a new version: https://github.com/graphql-go/graphql/releases/tag/v0.7.9, which includes this PR, thanks a lot again 👍 cheers!

PR #530, opened 2020-02-25
Add Subscribe function to make easier to create Subscription Handlers

racerxdl:
I added Subscribe function to Object / Fields Definitions so its easier to make GraphQL Subscription Handlers. I only added the field so it wouldn't break compatibility and if you don't use it, it wont make a difference.

So for example in https://github.com/racerxdl/go-subscription-handler I did:

```go
var rootSubscriptions = graphql.ObjectConfig{
    Name: "RootSubscriptions",
    Fields: graphql.Fields{
        "serverTime": &graphql.Field{
            Type: graphql.String,
                err := subhandler.Subscribe(p.Context, "serverTime")

                if p.Source != nil {
                    // We received a event data
                    v, ok := p.Source.(map[string]interface{})
                    if ok && v["time"] != nil {
                        return v["time"], nil
                    }
                }

                // We didn't receive a event data, so resolve normally
                return time.Now().String(), err
            },
        },
    },
}
```

With that `Subscribe` field in the struct I could do:

```graphql
var rootSubscriptions = graphql.ObjectConfig{
    Name: "RootSubscriptions",
    Fields: graphql.Fields{
        "serverTime": &graphql.Field{
            Type: graphql.String,
            Subscribe: func(p graphql.ResolveParams) error {
                return subhandler.Subscribe(p.Context, "serverTime")
            },
                if p.Source != nil {
                    // We received a event data
                    v, ok := p.Source.(map[string]interface{})
                    if ok && v["time"] != nil {
                        return v["time"], nil
                    }
                }

                // We didn't receive a event data, so resolve normally
                return time.Now().String(), err
            },
        },
    },
}
```

That should help with #242

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -339,6 +339,7 @@ type Object struct {
 	PrivateName        string `json:"name"`
 	PrivateDescription string `json:"description"`
 	IsTypeOf           IsTypeOfFn
+	Subscribe          SubscribeFn
 
 	typeConfig            ObjectConfig
 	initialisedFields     bool
@@ -368,12 +369,15 @@ type IsTypeOfFn func(p IsTypeOfParams) bool
 
 type InterfacesThunk func() []*Interface
 
+type SubscribeFn func(p ResolveParams) error
+
 type ObjectConfig struct {
 	Name        string      `json:"name"`
 	Interfaces  interface{} `json:"interfaces"`
 	Fields      interface{} `json:"fields"`
 	IsTypeOf    IsTypeOfFn  `json:"isTypeOf"`
 	Description string      `json:"description"`
+	Subscribe   SubscribeFn `json:"-"`
 }
 
 type FieldsThunk func() Fields
@@ -396,6 +400,7 @@ func NewObject(config ObjectConfig) *Object {
 	objectType.PrivateDescription = config.Description
 	objectType.IsTypeOf = config.IsTypeOf
 	objectType.typeConfig = config
+	objectType.Subscribe = config.Subscribe
 
 	return objectType
 }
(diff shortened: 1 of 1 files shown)


racerxdl:
@chris-ramon that's a quick hack to be able to do better subscriptions. There are other PRs about it  (like #495 ) but this one is pretty simple and fast to review. I just add a Subscription parameter so the handler can use it and have better code.

nic:
Nice job!! Any prediction of when it will be accept/merged?

jeduardocosta:
I'm also interested in this change. Any news?

remorses:
Instead of using a different subscribe functions we should use a channel to send events and return this channel as resolver return value

Gqlgen and graphql gophers already do this, 

I don’t like the approach of this pr, it is not intuitive and difficult to integrate with pubsub libs

definition.go
```diff
@@ -368,12 +369,15 @@ type IsTypeOfFn func(p IsTypeOfParams) bool
 
 type InterfacesThunk func() []*Interface
 
+type SubscribeFn func(p ResolveParams) error
```
bhoriuchi, line 372: this is the same function signature as a FieldResolveFn, any reason not to just use that?

bhoriuchi:
> Instead of using a different subscribe functions we should use a channel to send events and return this channel as resolver return value
> 
> Gqlgen and graphql gophers already do this,
> 
> I don’t like the approach of this pr, it is not intuitive and difficult to integrate with pubsub libs

I gave a longer answer in my PR but the short version is that this project follows the reference implementation which has this method https://github.com/graphql/graphql-js/blob/master/src/type/definition.js#L1001 . It it also not difficult to integrate into pubsub libs if you implement the Subscribe handler https://github.com/graphql/graphql-js/blob/master/src/subscription/subscribe.js#L63

remorses:
ok i think we can close this in favour of #495 that actually implements subscriptions

definition.go
exapsy: Wouldn't it be more user friendly to assign this under a name which includes "Subscription" in it? Like, if you wanna search for the particular subscription handler for subscription, its easy to identify it. Could be wrong.

racerxdl:
I agree @remorses , Closing in favor of https://github.com/graphql-go/graphql/pull/495

PR #536, opened 2020-04-08
Fix support for null literals

mpenick:
This builds on an existing PR: https://github.com/graphql-go/graphql/pull/401. It fixes issues found in that PR (it works for enums, objects, and lists values) and adds tests.

Externally, this uses `nil` instead of a special value/type i.e. `NullValue`. Internally, it differentiates between "not present" and "null" using a special `nullValue` type.

Give it a try and let me know what you think.


mpenick:
This also fixes issues with the incorrect element index being used in errors.

jorgebay:
We've been using this fix for the past week and works perfectly. The fix is straightforward and provides good coverage.

Is there anything I can do to help land this?

jeduardocosta:
Is this project still active? I've seen some pull requests with no answers, and that worries me.

bhoriuchi:
> Is this project still active? I've seen some pull requests with no answers, and that worries me.

It is but @chris-ramon is the only one approving/merging PRs. He seems to approve the straightforward ones every few months or so.

remorses:
I hoped this pr could resolve the #546 issue, unfortunately it gives the same error:
```
Cannot return null for non-nullable field Nested.field.
```
I think the problem is that go-graphql can't tell if the returned nil is null or not and tries to resolve the fields anyway

rules.go
```diff
@@ -1762,7 +1760,12 @@ func isValidLiteralValue(ttype Input, valueAST ast.Value) (bool, []string) {
 		if valueAST, ok := valueAST.(*ast.ListValue); ok {
 			messagesReduce := []string{}
 			for idx, value := range valueAST.Values {
-				_, messages := isValidLiteralValue(itemType, value)
+				var messages []string
+				if value.GetKind() == kinds.NullValue {
+					messages = []string{"Unexpected null literal."}
```
ddebrunner, line 1765: @mpenick Is there a reason for this disallowing null in a list? There's even a test for it so it doesn't seem to be an accident.

Section 2.9.7 seems to allow nulls in a list.
https://spec.graphql.org/June2018/#ListValue

vyj7:
@chris-cp  Can we merge this PR? As per GraphQL specs we allow "Null" in the list

krasish:
We also have a scenario with this issue, can we please fix it 🙏🙏

chrisfrank:
Just wanted to report that we merged this fix into our (currently private) graphql-go fork at NYT about six months ago, and have been running it in production ever since. If there's anything I can do to help this PR land here upstream, I'd be really happy to!

PR #540, opened 2020-04-25
return nil schema on error

nasermirzaei89:
Golang methods shouldn't return zero value structs on error. So, we should return `*Schema` on the `NewSchema` method.

Why:
https://groups.google.com/d/msg/golang-nuts/4yBldwsjZwo/sFcz1KVhU3wJ


bhoriuchi:
This will break a lot of people's code. So while I agree that it should be done, it should probably be done so in a major release.

nasermirzaei89:
I agree with merge it for a major release.

PR #561, opened 2020-08-03
fix: error detection

CxdInitial:
should check nullability of `gt.err` instead of local variable `err` which was checked before


CxdInitial:
BTW, `fieldConfig.Type` should be of type `Object` or `InputObject`, just checking nullability is not enough.

PR #573, opened 2020-10-08
feature: show POST example how all apps do it

### LINKED ISSUE #420: POST Example
How I can send a request by POST-Method with the following shema?

Under examples can I find only GET-Requests!

```
{ 
  "query" :  "..." , 
  "operation" :  "..." , 
  "variables" :  {  "myVariable" :  "somevalue" ,  . . .  } 
}
```

with : 

```
curl -XPOST http://localhost:5000/xql -H 'Content-Type: application/json' -d \
'{
  "query": "mutation tokenAuth($email:String, $password:String)",
  "variables": "{\"email\": \"emai@domain.com\", \"password\": \"password\"}"
}'
```

currently have I a error: "Syntax Error GraphQL request (2:3) Expected Name, found String \"query\...."

What's my problem? :)

maapteh:
I decided it would be nice to have an example with POST. Since it's more common. I also updated the playground which offers more options. So:

- have the TODO app working with complete working UPDATED playground which also has autocomplete (ctrl+space) on everything possible
- have app working in POST mode, with GET you show all given arguments which is not good to do on server level (security)

closes #420

Screenshot 2020-10-08 at 17 29 44

credits to @atombender who showed an example :)


chris-ramon:
@maapteh thanks a lot for working on this PR 👍 –– very interesting improvement for the `examples/todo`.

I think GraphQL Playground is very useful tool, but I think the `examples/todo` aims to demonstrate how to use `graphql-go/graphql` as a HTTP endpoint & how to integrate it with a Web App.

But I definitely understand your motivation to improve the `examples/todo` to show how to use `graphql-go/graphql` via `HTTP POST` and integrate it with `GraphQL Playground`.

For this type of use cases I would recommend to use `graphql-go/handler` which handles both HTTP verbs: [`GET`](https://github.com/graphql-go/handler/blob/c1267356e6539d975e83bbea383c2bbb861dce5b/handler.go#L68) & [`POST`](https://github.com/graphql-go/handler/blob/c1267356e6539d975e83bbea383c2bbb861dce5b/handler.go#L99).

Using [`graphql-go/handler`](https://github.com/graphql-go/handler) and the `examples/todo` schema, both can be integrated to have `GraphQL Playground` working, Eg:

```go
package main

import (
	"log"
	"net/http"

	"github.com/graphql-go/graphql/examples/todo/schema"
	"github.com/graphql-go/handler"
)

func main() {
	h := handler.New(&handler.Config{
		Schema:     &schema.TodoSchema,
		Pretty:     true,
		GraphiQL:   false,
		Playground: true,
	})

	http.Handle("/graphql", h)
	log.Println("server running on port :8080")
}
```

chris-ramon:
I went ahead and sent a [PR#574](https://github.com/graphql-go/graphql/pull/574) for exposing the `TodoSchema` so it can be use as shown above.

chris-ramon:
Closing this one in favor of: https://github.com/graphql-go/graphql/pull/574

maapteh:
Hi @chris-ramon i think you missed my POST only remark. In GraphQL frontend implementations all clients use POST. And GET is wrong because of two things: url length max and security. Thats why i wanted to make a POST only server and simply not support GET.

"In GraphQL applications, client requests are all HTTP POST requests and completely bypass the standard HTTP caching mechanism."

So if we still want to show the curl example we then do locally like:
```
curl \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{ "query": "{ todoList { text } }" }' \
  http://localhost:8080/graphql
```

By the way nice playground is already integrated. I missed that and maybe its then good to whipe it in your examples?

I didn't see option to turn introspection ON or OFF. Because on production you dont want to show your whole schema to the world.

PR #574, opened 2020-10-10
examples/todo: extracts TodoSchema

chris-ramon:
#### Overview

- examples/todo: extracts todo schema so it can be re-use for demo purposes.

#### Test plan

- Tested that `examples/todo` is working as expected:

```bash
(todo)-> go run main.go 
Now server is running on port 8080
Get single todo: curl -g 'http://localhost:8080/graphql?query={todo(id:"b"){id,text,done}}'
Create new todo: curl -g 'http://localhost:8080/graphql?query=mutation+_{createTodo(text:"My+new+todo"){id,text,done}}'
Update todo: curl -g 'http://localhost:8080/graphql?query=mutation+_{updateTodo(id:"a",done:true){id,text,done}}'
Load todo list: curl -g 'http://localhost:8080/graphql?query={todoList{id,text,done}}'
Access the web app via browser at 'http://localhost:8080'
```

```bash
(graphql)-> curl -g 'http://localhost:8080/graphql?query={todo(id:"b"){id,text,done}}'
{"data":{"todo":{"done":false,"id":"b","text":"This is the most important"}}}
```

```bash
(graphql)-> curl -g 'http://localhost:8080/graphql?query=mutation+_{createTodo(text:"My+new+todo"){id,text,done}}'
{"data":{"createTodo":{"done":false,"id":"qtVJmbwj","text":"My new todo"}}}
```

```bash
(graphql)-> curl -g 'http://localhost:8080/graphql?query=mutation+_{updateTodo(id:"a",done:true){id,text,done}}'
{"data":{"updateTodo":{"done":true,"id":"a","text":"A todo not to forget"}}}
```

```bash
(graphql)-> curl -g 'http://localhost:8080/graphql?query={todoList{id,text,done}}'
{"data":{"todoList":[{"done":true,"id":"a","text":"A todo not to forget"},{"done":false,"id":"b","text":"This is the most important"},{"done":false,"id":"c","text":"Please do this or else"},{"done":false,"id":"qtVJmbwj","text":"My new todo"}]}}
```

***

- Tested that `TodoSchema` works as expected using `graphql-go/handler`:

```go
package main

import (
	"log"
	"net/http"

	"github.com/graphql-go/graphql/examples/todo/schema"
	"github.com/graphql-go/handler"
)

func main() {
	h := handler.New(&handler.Config{
		Schema:     &schema.TodoSchema,
		Pretty:     true,
		GraphiQL:   false,
		Playground: true,
	})

	http.Handle("/graphql", h)
	log.Println("server running on port :8080")
}
```

### DIFF
--- a/examples/todo/main.go
+++ b/examples/todo/main.go
@@ -8,191 +8,17 @@ import (
 	"time"
 
 	"github.com/graphql-go/graphql"
+	"github.com/graphql-go/graphql/examples/todo/schema"
 )
 
-type Todo struct {
-	ID   string `json:"id"`
-	Text string `json:"text"`
-	Done bool   `json:"done"`
-}
-
-var TodoList []Todo
-var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
-
-func RandStringRunes(n int) string {
-	b := make([]rune, n)
-	for i := range b {
-		b[i] = letterRunes[rand.Intn(len(letterRunes))]
-	}
-	return string(b)
-}
-
 func init() {
-	todo1 := Todo{ID: "a", Text: "A todo not to forget", Done: false}
-	todo2 := Todo{ID: "b", Text: "This is the most important", Done: false}
-	todo3 := Todo{ID: "c", Text: "Please do this or else", Done: false}
-	TodoList = append(TodoList, todo1, todo2, todo3)
+	todo1 := schema.Todo{ID: "a", Text: "A todo not to forget", Done: false}
+	todo2 := schema.Todo{ID: "b", Text: "This is the most important", Done: false}
+	todo3 := schema.Todo{ID: "c", Text: "Please do this or else", Done: false}
+	schema.TodoList = append(schema.TodoList, todo1, todo2, todo3)
 
 	rand.Seed(time.Now().UnixNano())
 }
-
-// define custom GraphQL ObjectType `todoType` for our Golang struct `Todo`
-// Note that
-// - the fields in our todoType maps with the json tags for the fields in our struct
-// - the field type matches the field type in our struct
-var todoType = graphql.NewObject(graphql.ObjectConfig{
-	Name: "Todo",
-	Fields: graphql.Fields{
-		"id": &graphql.Field{
-			Type: graphql.String,
-		},
-		"text": &graphql.Field{
-			Type: graphql.String,
-		},
-		"done": &graphql.Field{
-			Type: graphql.Boolean,
-		},
-	},
-})
-
-// root mutation
-var rootMutation = graphql.NewObject(graphql.ObjectConfig{
-	Name: "RootMutation",
-	Fields: graphql.Fields{
-		/*
-			curl -g 'http://localhost:8080/graphql?query=mutation+_{createTodo(text:"My+new+todo"){id,text,done}}'
-		*/
-		"createTodo": &graphql.Field{
-			Type:        todoType, // the return type for this field
-			Description: "Create new todo",
-			Args: graphql.FieldConfigArgument{
-				"text": &graphql.ArgumentConfig{
-					Type: graphql.NewNonNull(graphql.String),
-				},
-			},
-			Resolve: func(params graphql.ResolveParams) (interface{}, error) {
-
-				// marshall and cast the argument value
-				text, _ := params.Args["text"].(string)
-
-				// figure out new id
-				newID := RandStringRunes(8)
-
-				// perform mutation operation here
-				// for e.g. create a Todo and save to DB.
-				newTodo := Todo{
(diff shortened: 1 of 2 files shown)


maapteh:
Why would you want to add `result := executeQuery(r.URL.Query().Get("query"), schema.TodoSchema` still?

Just update your curl options for this demo:

```
curl \
  -X POST \
  -H "Content-Type: application/json" \
  --data '{ "query": "{ todoList { id text } }" }' \
  http://localhost:8080/graphql
```

Then you dont open a GET option to get the results. All web clients use POST. By giving the options to GET it as well its superweird and users will think they can integrate like this as well.

examples/todo/main.go
```diff
@@ -207,7 +33,7 @@ func executeQuery(query string, schema graphql.Schema) *graphql.Result {
 
 func main() {
 	http.HandleFunc("/graphql", func(w http.ResponseWriter, r *http.Request) {
-		result := executeQuery(r.URL.Query().Get("query"), schema)
+		result := executeQuery(r.URL.Query().Get("query"), schema.TodoSchema)
 		json.NewEncoder(w).Encode(result)
 	})
```
maapteh, line 38: this is not utilising your sample given:
```
	h := handler.New(&handler.Config{
		Schema:     &schema.TodoSchema,
		Pretty:     true,
		GraphiQL:   false,
		Playground: true,
	})

	http.Handle("/graphql", h)
	log.Println("server running on port :8080")
```

why not take this static out of the example? and why still give GET example. Bu showcasing POST you close the open Issue and its shown by example it can do it :)

chris-ramon:
> Why would you want to add `result := executeQuery(r.URL.Query().Get("query"), schema.TodoSchema` still?

@maapteh FMHO we should keep the usage of `GET` just for simplicity, the usage of the the correct **HTTP verb** should rely on lib users, if a secure protocol is a requirement then a proper secure protocol should be implemented like **HTTPS**.


***


> 
> Just update your curl options for this demo:
> 
> ```
> curl \
>   -X POST \
>   -H "Content-Type: application/json" \
>   --data '{ "query": "{ todoList { id text } }" }' \
>   http://localhost:8080/graphql
> ```
> 

Planning to send a PR for adding `examples/http-post` in order to show the usage of `HTTP POST`, and yes we can continue to use `curl` for consistency with `examples/todo` to show the usage of the endpoint.

***

> Then you dont open a GET option to get the results. All web clients use POST. By giving the options to GET it as well its superweird and users will think they can integrate like this as well. But maybe you like it to act like REST?

As mentioned above, FMHO lib users should decide which HTTP verb to use, we just use GET in the example for simplicity.

***

> 
> But nice work, i totally missed the option by default you can open playground and have POST working. This wasn't demonstrated in any example.

Yes, good call, I've created [`graphql-go/handler-examples/todo`](https://github.com/graphql-go/handler-examples/blob/master/todo/main.go) for the matter.

***

> Im wondering what happens if i turn it off (on production we turn off playground and introspection and deploy another instance with stuff on, which is only reachable internally).
> 
> Have a nice day

Using `graphql-go/handler` the GraphQL Playground can be disable via the `Playground: false` option, disabling it does not interfere the correct execution of the `graphql-go/handler` itself.

examples/todo/main.go
chris-ramon: @maapteh yes, the given example just shows how to connect `handle.New` with `schema.TodoSchema`, in order to show this kind of examples I've created a new repository [`graphql-go/handler-examples`](https://github.com/graphql-go/handler-examples) which we can use to point lib users to in order to show `graphql-go/handler` related examples.

I think is ok to keep the usage of `GET` in `examples/todo` for simplicity of the example itself.

In order to showcase the usage of `POST`, planning to send  a PR for adding a new example `examples/http-post` which will be based from your [PR#573](https://github.com/graphql-go/graphql/pull/573).

chris-ramon:
> Isnt it possible to have an example with POST only server? For example in another setup i learned i could do:
> 
> ```
> 	r := gin.New()
> 	
> 	r.GET("/", playgroundHandler())
> 	r.POST("/graphql", graphqlHandler())
> 	r.GET("/.well-known/server-health", healthHandler())
> ```
> 
> and choose myself simple

Yes, planning to send a PR for adding a new example: `examples/http-post` for the matter after margining this PR.

examples/todo/main.go
chris-ramon: ^ Related PR for adding `examples/http-post`: https://github.com/graphql-go/graphql/pull/575

chris-ramon:
> Yes, planning to send a PR for adding a new example: `examples/http-post` for the matter after margining this PR.

^ Related PR for adding `examples/http-post`: https://github.com/graphql-go/graphql/pull/575

maapteh:
Nice work Chris-Ramon!

PR #587, opened 2020-12-08
Fix quotation during printing StringValue node

### LINKED ISSUE #586: Incorrect quotation when printing StringValue node
Hi,

I noticed an issue with printing `StringValue` value ([./language/printer/printer.go#L372](https://github.com/graphql-go/graphql/blob/master/language/printer/printer.go#L372)). The quotation of the string value is incorrect. If the input StringValue node contains escaped double quote characters (`\"`), then the output string from printer drops proper character escaping.

### Steps to reproduce

If we get a following query and parse it to `*ast.Document`, and then print it to a string:
```go
queryAst := `query { foo(jsonStr: "{\"foo\": \"bar\"}") }`
astDoc, _ := parser.Parse(parser.ParseParams{
		Source: query,
		Options: parser.ParseOptions{
			NoLocation: true,
		},
	})
result := printer.Print(astDoc)
```

### Expected result

We should expect the result string equal to:

```graphql
{
  foo(jsonStr: "{\"foo\": \"bar\"}")
}
```

### Actual result

However, what we get instead is:

```graphql
{
  foo(jsonStr: "{"foo": "bar"}")
}
```

Which makes it incorrect query.

I will provide a pull request with a unit test case and fix for that issue.

Cheers!

pkosiec:
Fixes #586.

See the details of the bug in the linked issue.


pkosiec:
Hi @chris-ramon, any news regarding this PR? Can you review it please? It's just 16 lines. Thanks!

chris-ramon (approved):
LGTM 👍 –– Thanks a lot @pkosiec 🚢

PR #589, opened 2020-12-14
BindFields: added default fields tag

gstarikov:
hi. i didn't understand why it necessary to specify json tags for every field. so i added default values for BindFields function.

### DIFF
--- a/examples/sql-nullstring/main.go
+++ b/examples/sql-nullstring/main.go
@@ -192,6 +192,9 @@ query {
 		log.Fatal(r1)
 	}
 	b1, err := json.MarshalIndent(r1, "", "  ")
+	if err != nil {
+		log.Fatal(err)
+	}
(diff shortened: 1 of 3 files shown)


williamhaley:
Agreed. I ran into this same issue today. The current behavior that requires specifying a tag is a bit intuitive. I think this PR makes a lot of sense. Is there anything I or others can do to help push this along?

FWIW, I tried a workaround in the meantime...

```
graphql.BindFields(CloneAndAddTags(myType{}))
```

With `CloneAndAddTags` I used reflection to dynamically clone the source struct `Type` and set the appropriate `json` tags so that `BindFields` could understand, but that failed. `BindFields` generates a resolver that [invokes `extractTag`](https://github.com/graphql-go/graphql/blob/62a7bb0a9839309631e20fc4ed032cbf00f88544/util.go#L133) so my hack was doomed.

PR #591, opened 2021-01-07
Updated the link to the latest pkg.go.dev

ujjwal-kr:
As in https://blog.golang.org/godoc.org-redirect states that the new documentation link will be redirected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Supports: queries, mutations & subscriptions.
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


chris-ramon (approved):
LGTM 👍 –– Thanks a lot @ujjwal-kr

Since traffic will be redirected from `godoc.org` to `pkg.go.dev` early this year, let's already replace the link.

Source: https://github.com/golang/go/issues/43178#issue-766770804

PR #605, opened 2021-06-09
modify make token by byte position, not rune position at readName

nakatamixi:
Hi!

I tried to parse [SDL with multibyte comment](https://github.com/graphql-go/graphql/compare/master...nakatamixi:fix/multi-byte-parse?expand=1#diff-eb480293457502591c6bfe13bf309a4fe6ab9b6c3052f4bedaae1a8c04a105a8R503-R506) , and failed.

I think, the cause is that the start and end positions of the token created in `readName` were created as the rune position.
If the comment contains a multibyte character string, the start and end of the token shifts from the actual byte position, and it seems that the token is not parsed correctly.

Please review this.

Thanks!


saranrapjs:
Confirming that this bug still exists on `master`, and this PR fixes the bug!

Issue #65, opened 2015-11-06
type of ResolveInfo.RootValue should be interface{}


bsr203:
this was discussed by few at https://github.com/graphql-go/graphql/pull/25

this avoid an extra casting

sogko:
Hi @bsr203 
Thanks for bring this up and keeping it from getting buried 👍🏻

From @pyros2097 comments in #25 
> I was just going to post an issue on the inconsistent API passing root value,
In the Executor params you specify Root as an interface{} here https://github.com/chris-ramon/graphql-go/blob/master/executor/executor.go#L15
But in the GraphqlParams you specify it as a map[string]interface{} here https://github.com/chris-ramon/graphql-go/blob/master/graphql.go#L15
>Maybe the RootObject should be an interface because anyway we are going to attach structs to it like session data or some other context which we will be typecasting to anyway.

I think the main take-away from that PR is that `RootValue` can be used to provide context for each GraphQL request (see https://github.com/graphql/graphql-js/issues/56). But currently, `RootValue` type is inconsistent.

```go
package graphql

// passed into `graphql.Graphql(...)`
type Params struct {
	RootObject     map[string]interface{}
}

type ExecuteParams struct {
	Root          interface{}
}

type BuildExecutionCtxParams struct {
	Root          interface{}
}

// finally available as RootValue in ResolveInfo through GQLFRParams 
type ResolveInfo struct {
	RootValue      interface{}
}
```

Along the way, RootObject gets renamed as Root, and finally RootValue.
In addition to that, to add to the confusion, in `FieldResolveFn`, RootValue is made available through `GQLFRParams` as both `p.Source` and `p.Info.RootValue`. (i.e. `p.Source === p.Info.RootValue`)

It would be great to have a consistent name for `RootValue`.

While I'm leaning towards updating RootObject/RootValue to `map[string]interface{}`, 
I would like to hear if anyone think it should be `interface{}` instead? (Or another type, perhaps `context`)?

Changing it `map[string]interface` may not necessarily remove extra casting; you may still need to cast the `interface{}` value stored in the map most of the time anyway. But it does allow you to store multiple context values in a map and make it available in `FieldResolveFn`.

Discussion welcomed!
Cheers guys!

bsr203:
It does avoid one casting though

```
Resolve: func(p gt.GQLFRParams) interface{} {
            rv := p.Info.RootValue.(map[string]interface{})
            ctx := rv["ctx"].(context.Context)
...
}
```

[...]

PR #608, opened 2021-08-03
fix: issue #65: type of ResolveInfo.RootValue should be interface{}

chirino:
changes type of ResolveInfo.RootValue ot be interface{}

I would like to set the RootObject to a struct pointer that my resolvers would rather deal with.


chris-ramon:
@chirino thanks for filing a PR — I don't think we should be changing this end-user type: `graphql.Params.RootObject` because it would be an API breaking-change.

An alternative strategy would be to have a small wrapper around your resolvers that consolidates `map[string]interface{}` to `interface{}` on your end ?

chirino:
To avoid the breaking change we could Introduce a new field and deprecate RootValue and ask folks to stop using RootValue.

chirino:
@chris-ramon what about this new version?

bhoriuchi:
RootValue is part of the spec, I don't think deprecating it is a good idea.

That said, will changing it to `interface{}` actually break it? `map[string]interface{}` can be assigned to an interface value no?

PR #622, opened 2022-01-14
Support thunks for Union.Types definitions

alex-lange:
In schemas, it is possible to define the Union type before the
definitions of the object types that are part of the union. This
change adds a "UnionTypesThunk" option when creating a new Union,
similar to the existing FieldsThunk and InterfacesThunk.

This more closely matches the interface in graphql-js:
  https://github.com/graphql/graphql-js/blob/47bd8c8897c72d3efc17ecb1599a95cee6bac5e8/src/type/definition.ts#L1307

It is a first step in closing out https://github.com/graphql-go/graphql/issues/624

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -796,15 +796,19 @@ type Union struct {
 	PrivateDescription string `json:"description"`
 	ResolveType        ResolveTypeFn
 
-	typeConfig    UnionConfig
-	types         []*Object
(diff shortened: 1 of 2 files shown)


chris-ramon (approved):
LGTM :+1: — Neat improvement @alex-lange, thanks a lot!

Issue #629, opened 2022-04-28
null values in input are omitted in ResolveParams.Args


Youngbae-Jeon:
When I request a mutation to my graphql server(using graphql-go) with null values in input arguments,
the null values don't come in ResolveParams.Args
```graphql
mutation Update($id: Int!, $input: UpdateInput!) {
    update(id=$id, input=$input)
}
```
with variable...
```javascript
{
    id: 123,
    input: {
        title: "TITLE"
        description: null
    }
}
```
Property description does not come in mutation resolver function
```go
// Resolver for mutation update
func ResolveUpdate(p graphql.ResolveParams) (any, error) {
    fmt.Println("*** Args:", p.Args)
    // -> *** Args: map[id:123 input:map[title:TITLE]]
}
```

PS.
When I write graphql server with graphql-js (javascript version of graphql library),
null values comes into the resolver implementation function normally.

PR #631, opened 2022-05-06
fix: input object with a nullable field set to null was being removed…

chirino:
… from the `Args` map given to a resolver.

### DIFF
--- a/values.go
+++ b/values.go
@@ -156,11 +156,12 @@ func coerceValue(ttype Input, value interface{}) interface{} {
 		}
 
 		for name, field := range ttype.Fields() {
-			fieldValue := coerceValue(field.Type, valueMap[name])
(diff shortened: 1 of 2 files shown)


ki4jnq:
:+1: Running into this on our project, too. Seems there's an existing issue as well, https://github.com/graphql-go/graphql/issues/629

adaam2:
Can this be merged as we are also running into this?

kindermax:
Hi, what chances are that this PR could be merged?

adaam2:
@kindermax As far as I can remember, I don't think that this PR is a complete implementation of nulls - we ended up forking this repo and doing it ourselves but I've since left that company so not sure I have access anymore.

kindermax:
I see( Well for now I've overcome this by passing variables via context but it is a dirty hack.

PR #636, opened 2022-07-10
fix: introspection query would give results in random order.

chirino:
apply sorting on lists generated from maps.


aluzzardi:
Any plans on merging this?

PR #651, opened 2022-10-17
fix String/Name methods on the List type

dariuszkuc:
`String()` method should delegate to the `Name()` method (same as `NonNull`). Trying to get name of a list type was returning just the underlying list element type info.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -193,13 +193,12 @@ func GetNamed(ttype Type) Named {
 //
 // Example:
 //
-//    var OddType = new Scalar({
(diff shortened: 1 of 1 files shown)


dariuszkuc:
Looks like vscode autoformatted some comments - let me know if you want those changes reverted.

dariuszkuc:
@chris-ramon can you help with merging this PR?

chris-ramon (approved):
LGTM 👍 – Thanks a lot @dariuszkuc

Issue #492, opened 2019-07-01
Feature Request: Support Apollo Graphql Federation


StevenACoffman:
## Feature Request

Please consider supporting [Graphql Federation](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/).

We use hundreds of microservices, and a monolithic GraphQL server becomes an unacceptable development bottleneck and single point of failure, so it becomes necessary to divide the graph's implemention into separate parts. We tried schema stitching, but would prefer federation for three reasons:

1.    With federation, microservices all expose a proper part of the overall graph and can refer directly to types that live in other services, without the need to add foreign keys or superfluous relationship root fields to your schema.
2.    Federation is fully declarative and doesn't require any user code to be running in the gateway.
3.    Execution is efficient and predictable, because it relies on a query plan generated ahead of time with full knowledge of the overall query, instead of on runtime schema delegation that gets invoked as part of normal resolver-based execution.

https://www.apollographql.com/docs/apollo-server/federation/federation-spec/

bhoriuchi:
this seems outside the scope of this project.

StevenACoffman:
How so?

bhoriuchi:
Federation is an Apollo spec not a graphql spec. This library implements the graphql spec https://graphql.github.io/graphql-spec/June2018/ which has no mention of federation.

bhoriuchi:
Actually I apologize. I realize you are just asking this library to support the additions not federation itself. I believe the additions should be in scope

StevenACoffman:
No worries, I clarified my request. Sorry for not being more clear from the beginning.

0xR:
You could implement federation in a separate library. See for example: https://github.com/0xR/graphql-transform-federation. It takes an existing schema and with some configuration it can add federation support. Note that that is a NodeJS implementation, but I'm sure it could be implemented in Go as well.

If you're ok with running NodeJS you can add federation support to your Go service by using graphql-transform-federation as a middleware service. It can take a remote schema as input as well.

itmayziii:
@chris-cp 

Is there interest in this library supporting this? I've read the [federation spec](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/) a couple of times now and it seems at a high level this library would need to support:

[...]

PR #652, opened 2022-10-17
feat: initial suport for applied directives

dariuszkuc:
Update type definitions to expose `AppliedDirective`s. **This initial implementation only exposes applied directive as metadata and does not modify runtime wiring to process them**. Exposing directives is a pre-req to enable support for [Apollo Federation](https://www.apollographql.com/docs/federation/) (related: https://github.com/graphql-go/graphql/issues/492).

Following the pattern from [`graphql-java`](https://github.com/graphql-java/graphql-java) and [`HotChocolate`](https://github.com/ChilliCream/hotchocolate) where they make distinction between directive type definition (i.e. `directive @foo on FIELD_DEFINITION`, maps to existing `Directive struct` for backwards compatibility) and where directives are actually applied within the schema (i.e. `type Foo @bar { baz @qux }`, maps to new `AppliedDirective` struct`).

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -193,16 +193,16 @@ func GetNamed(ttype Type) Named {
 //
 // Example:
 //
-//    var OddType = new Scalar({
-//      name: 'Odd',
-//      serialize(value) {
-//        return value % 2 === 1 ? value : null;
-//      }
-//    });
-//
+//	var OddType = new Scalar({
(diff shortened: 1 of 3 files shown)


dariuszkuc:
@chris-ramon can you help with merging this PR?

PR #655, opened 2022-11-16
Expose ParseValue

victor-perov:
This PR marks ParseValue as a public method that allows it to be called directly. It is handy to get AST value as part of pre-processing / validation.


victor-perov:
Reopened as #656 as a cleaned version.

PR #672, opened 2023-05-28
fix: removed extra blank link between sections

adi-igit:


chris-ramon:
@adi-igit thanks for the PR but it does include redundant white space & blank line to the markdown formatting.

PR #673, opened 2023-07-20
fix postData's operationName on http-post example

qustavo:
As per graphql's documentation the field should be `operationName`[1]

1: https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#post


chris-ramon (approved):
LGTM 👍 – Thanks @qustavo

PR #708, opened 2025-01-03
Fix incorrect lexing and parsing of multi-byte comments

saranrapjs:
We recently ran into a bug where including the unicode em-dash character (`—`) lead to GraphQL queries with that character in the comments to fail to be parsed. Digging into the issue further, it looks like some of the existing handling for lexing comment strings as whitespace fail to account for multi-byte comments in how they count byte length of commented unicode characters.

This PR fixes the bug (by including the byte length of commented characters in the "runePosition" count that's used to fast forward whitespace), and adds a couple of tests in the `lexer` and `parser` packages which should illustrate the bug as it stands on `main`.

### DIFF
--- a/language/lexer/lexer.go
+++ b/language/lexer/lexer.go
@@ -625,7 +625,7 @@ func positionAfterWhitespace(body []byte, startPosition int) (position int, rune
 						// SourceCharacter but not LineTerminator
 						(code > 0x001F || code == 0x0009) && code != 0x000A && code != 0x000D {
 						position += n
-						runePosition++
+						runePosition += n
 						continue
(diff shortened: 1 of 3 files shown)


saranrapjs:
I'm realizing this addresses the same bug as this PR: https://github.com/graphql-go/graphql/pull/605

saranrapjs:
...I'm going to close this in favor of the other [PR](https://github.com/graphql-go/graphql/pull/605), which also addresses multi-byte character sequences in non-comment ranges (e.g. a BOM character) in addition to the comments bug. It also passes the tests I'd added here.

PR #716, opened 2025-03-28
README.md: Adds contribute back section

chris-ramon:
#### Details
- `README.md`: Adds contribute back section.

#### Test Plan
:heavy_check_mark: Tested that contribute back section renders as expected:

Screenshot from 2025-03-28 12-13-33

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ godoc: https://pkg.go.dev/github.com/graphql-go/graphql
 
 Friendly reminder links are available in case you would like to contribute back into our commitment with Go and open-source.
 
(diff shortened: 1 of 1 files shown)


chris-ramon:
@mentatbot I've noticed that your comment above could be simplify from:

**[ ] Wake on any new activity.**

_I always see new activity while actively working. When paused, I'll only respond when tagged unless this box is checked._


***

To, as a contributor it is clearer when to enable the checkbox:


**[x] Respond when tagged.**

_I always respond on new activity when tagged._

PR #730, opened 2025-10-29
fix(rules): match spec for defaulted required args

lbarthon:
This adds a new test, that fails on master, alongside with the correct changes for it to work as expected.

The validation was failing for arguments that are defined as required, but have a default value, preventing a non-required argument to override their value, while it should work as expected.


lbarthon:
Would love to land this, and get a new version tagged if possible @chris-ramon 🙏

PR #741, opened 2026-05-20
fix: terminate overlapping-fields rule on cyclic fragment spreads

### LINKED ISSUE #742: OverlappingFieldsCanBeMergedRule stack-overflows on cyclic fragment spreads
Validating a document that contains a cyclic fragment spread crashes the process with a stack overflow.

## Reproducer

```graphql
fragment A on Dog { ...B }
fragment B on Dog { ...A }
{ ...A }
```

Running this through `graphql.ValidateDocument` against any schema produces:

```
runtime: goroutine stack exceeds 1000000000-byte limit
fatal error: stack overflow
```

The stack points at `collectConflictsBetweenFieldsAndFragment` in `rules_overlapping_fields_can_be_merged.go` recursing on itself.

## Impact

Any server using this library can be crashed by a small malformed query. `NoFragmentCyclesRule` correctly reports the cycle as a validation error, but `OverlappingFieldsCanBeMergedRule` runs in parallel via the visitor and crashes the process before that result is returned.

## Fix

PR #741 proposes a fix.

mennatnaga:
The overlapping-fields validator infinitely recurses on cyclic fragment spreads.                         
                                                                                                         
## How to reproduce                                                                                     
                                                                                                         
```graphql                                                                                             
  fragment A on Dog { ...B }                                                                             
  fragment B on Dog { ...A }                                                                             
  { ...A }                                                                                               
```                                                                                                
Before this patch: runtime: goroutine stack exceeds 1000000000-byte limit; fatal error: stack overflow inside collectConflictsBetweenFieldsAndFragment at rules_overlapping_fields_can_be_merged.go:209.      
                                                                                                         
After: validation returns Cannot spread fragment "A" within itself via B. from NoFragmentCyclesRule.
                                                                                                         
## Root cause                                                                           
                                          
collectConflictsBetweenFieldsAndFragment recurses through nested spreads without memoization. Its sibling collectConflictsBetweenFragments already has the equivalent guard via rule.comparedSet; this one was missing it.                     
                                                                                                         
## Changes                                                                                                
                                     
  - New fieldsAndFragmentSet memo keyed by (*fieldsAndFragmentNames, fragmentName, areMutuallyExclusive), same Has/Add shape as the existing pairSet.                                         
  - Wired into the rule constructor and struct alongside comparedSet and cacheMap.
  - Function-entry check skips repeat triples.
  - Added a fieldsInfo == fieldsInfo2 guard so a fragment is not compared against itself.
                                                                                                         
## Test                                        
                                                                                                         
TestValidate_OverlappingFieldsCanBeMerged_CyclicFragmentTerminates runs the rule inside a goroutine with a 2s deadline. Fails on the unpatched code (the goroutine never finishes), passes after the fix.
Existing OverlappingFields* tests still pass.
## Closes 
#742 

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation no longer hangs on documents with cyclic fragment spreads; overlapping-fields checks terminate reliably and run faster thanks to per-pass memoization.

* **Tests**
  * Added a test ensuring validation completes for cyclic fragment structures.

[Review Change Stack](https://app.coderabbit.ai/change-stack/graphql-go/graphql/pull/741?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/rules_overlapping_fields_can_be_merged.go
+++ b/rules_overlapping_fields_can_be_merged.go
@@ -50,6 +50,12 @@ func OverlappingFieldsCanBeMergedRule(context *ValidationContext) *ValidationRul
 	// dramatically improve the performance of this validator.
 	comparedSet := newPairSet()
 
+	// Memoizes (fields, fragment, areMutuallyExclusive) triples that have
+	// been compared by collectConflictsBetweenFieldsAndFragment in this
+	// validation pass. Subsequent calls on the same triple return without
+	// re-walking the fragment.
+	comparedFieldsAndFragmentSet := newFieldsAndFragmentSet()
+
 	// A cache for the "field map" and list of fragment names found in any given
 	// selection set. Selection sets may be asked for this information multiple
 	// times, so this improves the performance of this validator.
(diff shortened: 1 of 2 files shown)


chris-ramon (approved):
LGTM 👍 🚢 — Thanks a lot @mennatnaga

Issue #731, opened 2025-11-06
`definition.go`: Increase test coverage up to 90%.


chris-ramon:
#### Current

- Currently the test coverage of `github.com/graphql-go/graphql/definition.go` is at `(84.2%)`.

#### Expected

- We want to increase it up to 90%.

chris-ramon:
Closed via: https://github.com/graphql-go/graphql/pull/743

PR #743, opened 2026-06-06
`definition`: adds unit tests to reach 100% test coverage

chris-ramon:
#### Details
- Issue: https://github.com/graphql-go/graphql/issues/731.
- Adds unit tests via new file: `definition_extended_test.go`.
- Removes dead code found:
```
Final state of definition.go:485-493:
if iface.ResolveType != nil {
    err = invariantf(
        iface.ResolveType != nil,
        `Interface Type %v does not provide a "resolveType" function `+
            `and implementing Type %v does not provide a "isTypeOf" `+
            `function. There is no way to resolve this implementing type `+
            `during execution.`, iface, ttype,
    )
}
The only change from the original was removing the unreachable if err != nil { return } (the invariantf condition iface.ResolveType != nil can never be false inside the if iface.ResolveType != nil block, so err is always nil). The logic is identical — no behavior change.
```

***

## Summary by CodeRabbit

* **Bug Fixes**
  * Interface validation no longer aborts processing when a validation check fails; processing continues and the interface is preserved.

* **Tests**
  * Large new test suite covering GraphQL type definitions and behaviors for scalars, objects, interfaces, unions, enums, input objects, wrappers, and related serialization/parsing.

### DIFF
--- a/definition.go
+++ b/definition.go
@@ -490,9 +490,6 @@ func defineInterfaces(ttype *Object, interfaces []*Interface) ([]*Interface, err
 					`function. There is no way to resolve this implementing type `+
 					`during execution.`, iface, ttype,
 			)
-			if err != nil {
-				return ifaces, err
-			}
 		}
 		ifaces = append(ifaces, iface)
 	}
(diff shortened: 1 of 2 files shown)


chris-ramon (commented):
LGTM 👍 🚢

PR #747, opened 2026-06-14
new plan abstract-field selections lazily to avoid exponential blow-up

patrikhermansson:
After recent changes introducing plan caching led to big performance regressions, complex queries took seconds to plan. This is mostly a Claude generated PR so I don't take any credit for it but after benchmarking it brought planning time to existing baseline.

PlanQuery eagerly expanded every possible type of every Interface/Union field at plan time. Because an expanded type's sub-selection can contain further abstract fields, this lead to a big blow up in planing time for complex schemas.

What this changes is to actually make it lazy: planMergedFieldChildren leaves abstract alternatives unplanned, and completePlannedAbstractValue plans (and caches) the concrete type's sub-selection the first time that type is encountered at runtime, guarded by a mutex since fields resolve concurrently.

Plan time for abstract fields is now proportional to the types a request actually resolves rather than schema_size ^ depth. 

Also adding a plan_blowup_test.go: a regression guard (deeply-nested abstract selection must plan within a deadline), a correctness test, and a benchmark. Without the fix the guard does ~16^12 units of work and times out; with it, PlanQuery is ~constant-time (6 allocs/op) regardless of depth.

## Release Notes

* **Improvements**
  * Enhanced concurrent execution safety for abstract field planning through lazy initialization.
  * Optimized query planning for abstract types (Interface/Union) to defer evaluation until execution time, improving performance with complex nested selections.

* **Tests**
  * Added performance regression tests to prevent slowdowns with deeply nested abstract type selections.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -55,6 +55,10 @@ type executionContext struct {
 	VariableValues map[string]interface{}
 	Errors         []gqlerrors.FormattedError
 	Context        context.Context
+
+	// plan is set on the ExecutePlan path; it lets abstract fields plan
+	// their concrete-type sub-selections lazily at execute time.
+	plan *Plan
 }
 
 func buildExecutionContext(p buildExecutionCtxParams) (*executionContext, error) {
--- a/plan.go
+++ b/plan.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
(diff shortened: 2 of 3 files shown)


chris-ramon (approved):
LGTM 👍 🚢

